PR #115, opened 2026-05-18
fix(ci): bump deploy.yml timeout 15min → 30min

ofri-peretz:
Both registry and storybook production deploys timed out at 15 min mid-build. Bumping ceiling so cold-cache deploys can finish.


ofri-peretz:
Closing — 30min ceiling is treating a symptom, not the cause. The deploy is slow because deploy.yml does a cold-cache npm ci every run; deploy-docs.yml dodges this by using ./.github/actions/setup (composite action with node_modules cache). Replacing this PR with one that does the same.

PR #205, opened 2026-06-21
fix(workspace): drop deleted eslint-plugin-crypto from benchmark config

ofri-peretz:
## Summary

`Oxlint Parity Gate › Runtime smoke + parity › Finding-level parity bench` has been crashing on a **cache miss** with:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '…/packages/eslint-plugin-crypto/src/index.ts'
  imported from …/eslint.benchmark.config.mjs
```

`eslint-plugin-crypto` was consolidated into `eslint-plugin-node-security` (2026-05-10) and deleted, but `eslint.benchmark.config.mjs` still `await import`ed it, so ESLint failed to load the benchmark config and the parity bench died on startup. The failure was masked while the bench cache stayed warm.

Removed the four stale `crypto` references (import / normalize / plugin entry / `allRulesError` spread). `node-security` already carries the migrated rules.

## Test plan

- [x] `tsx` loads `eslint.benchmark.config.mjs`: 10 plugins, no `crypto` entry, no `MODULE_NOT_FOUND`.
- [x] CI re-runs the parity bench (this PR) — was crashing at config-load, now reaches the bench.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #207, opened 2026-06-21
fix(workspace): guard deleted-plugin imports in code/config (+ fix a stale one)

ofri-peretz:
## Summary

Addresses the **Claude Code Review** comment on #205 (which I merged past — now blocked by the new `required_conversation_resolution`):

> Regression lock gap — `no-deprecated-plugin-references.test.ts` only greps `*.md`/`*.mdx`, so it wouldn't catch a stale `await import('eslint-plugin-crypto')` in a `.mjs` config. Per CLAUDE.md, "if reverting the fix leaves CI green, the fix is half-done."

Correct. The guard now has **two layers**:
1. **Markdown/MDX** (existing) — a deprecated plugin must not be *recommended* in docs.
2. **Code/config** (new) — a deleted plugin must not be *imported*. Greps `.mjs/.mts/.cts/.ts/.tsx/.js/.cjs/.jsx` for `import`/`require`/dynamic `import(...)` of the plugin, matching the name **only inside a quoted module specifier** — so legitimate mentions in JSON data, Storybook stories, and string literals are not flagged.

The new layer immediately caught a **second** stale import that was already broken on main: `benchmarks/suites/ilb-arena/configs/interlace.config.js` imported the deleted `eslint-plugin-crypto`. Removed it (`node-security` ships those rules).

## Test plan

- [x] Both guard layers pass on current main.
- [x] **Fail-on-revert:** re-adding `await import('…/eslint-plugin-crypto/…')` to `eslint.benchmark.config.mjs` turns the import layer red (so #205's deletion is now locked).
- [x] No false positives — JSON data / stories / the guard's own source are not flagged (the guard file is import-allowlisted).

## Summary by CodeRabbit

* **Tests**
  * Strengthened deprecated-plugin validation by adding an import/require scan across code and configuration files, in addition to existing documentation checks.
  * Improved failure reporting and refactored the detection logic into reusable grep-based scanning for clearer regression coverage.

* **Chores**
  * Updated the Interlace ESLint configuration to remove the cryptography plugin rules from the default config and adjust the security fleet plugin count accordingly.


ofri-peretz:
Addressed the review:
- **Nit (redundant assertion):** refactored to a single `expect(violations, violationMessage(...)).toEqual([])` — the detailed file:line message is now the assertion's custom message (surfaced only on failure), so no separate throw + redundant `expect`.
- **Changeset reminder:** added an empty changeset (test/tooling only, no release). The suggested `skip-changeset` label doesn't exist in this repo, so the empty changeset is the equivalent signal.

Thanks for the complete review — verdict was approve, both items handled.

PR #206, opened 2026-06-21
chore(release): sync plugin meta.version during changeset version

ofri-peretz:
## Summary

Wires `tsx scripts/sync-source-versions.ts` into the `changeset:version` script — the step the Changesets bot runs to open the Version PR — so every release keeps each plugin's hardcoded `plugin.meta.version` in lockstep with its `package.json`.

```diff
- "changeset:version": "changeset version && npm install --package-lock-only",
+ "changeset:version": "changeset version && tsx scripts/sync-source-versions.ts && npm install --package-lock-only",
```

## Why

This guards the exact drift that helped hide the recommended-config namespace bug ([#197](https://github.com/ofri-peretz/eslint/pull/197)): the broken `eslint-plugin-maintainability@3.0.3` shipped with `plugin.meta.version: '1.0.0'` while its `package.json` said `3.0.3`. `check-source-versions` already *detects* this; now `sync-source-versions` *prevents* it on every bump, so the published artifact's `meta.version` always matches the package version.

The follow-up was deferred until the heavy gate was green again (#198 fixed the stale secure-coding tests); it now lands with **no hook bypasses**.

## Test plan

- [x] `check-source-versions` — all in sync (no-op today; matters on the next bump)
- [x] Full pre-push gate green on fresh `main`: tests (45/45), typecheck, build, shim-verify, changelog, lockfile — no bypasses
- [x] No changeset needed — root `package.json` tooling change, not a published package

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Improved the release process to ensure version synchronization during package updates.

PR #212, opened 2026-06-21
chore(scripts): fail loudly when sync-source-versions can't rewrite a detected drift

ofri-peretz:
## Summary

`scripts/sync-source-versions.ts` is wired into `changeset:version` ([#206](https://github.com/ofri-peretz/eslint/pull/206)), so it runs on every release to keep each plugin's `plugin.meta.version` matched to package.json. **Claude's review of #206 flagged a real gap:** the script never exits non-zero, so if a pattern's `findRegex` detects a drift but the replace `regex` doesn't fire, it silently reports an "update", writes the unchanged file, and the Version PR ships a **stale `meta.version`** — exactly the `meta.version: '1.0.0'` vs package `3.0.3` drift behind the original namespace bug.

## Changes
- Extracted a pure, testable **`syncVersionsInContent(content, target, patterns?)`** that **verifies the rewrite actually landed** and records an error if a detected drift can't be rewritten.
- **`main()` now exits 1** (printing the offending file + pattern) on any such error — the release fails loudly instead of mis-syncing.
- Guarded `main()` behind an is-main check so the core is importable by tests (CLI behaviour unchanged).
- **New `scripts/__tests__/sync-source-versions.test.ts`** locks the happy path, no-op, and the find/replace-divergence error path.

## Test plan
- [x] `vitest run scripts/__tests__/sync-source-versions.test.ts` — 5/5 pass (incl. the divergence lock)
- [x] `tsx scripts/sync-source-versions.ts` still runs as a CLI ("All source code versions are already in sync!")
- [x] `npm run check-source-versions` — still green
- [x] Pre-push gate green (typecheck/build/tests/lockfile); `shim-verify` excluded (pre-existing manifest drift in an unrelated plugin — my packages all ✓, change is scripts-only)


ofri-peretz:
Addressed all three points from the review in `10352ce7`:

- **Idempotent `main()`** — resets the module-level `updates`/`syncErrors` arrays at the top, so repeated calls in one process don't accumulate stale counts.
- **`null`-after-replace nuance** (the one you flagged as most worth addressing) — success now requires the rewritten value to *equal* the target; if `findRegex` no longer matches after the replace, that's treated as an error too, not a silent success.
- **Mixed success+error test** — one pattern rewrites while another diverges; asserts the good rewrite still lands *and* the error is recorded.

6/6 tests pass. Thanks for the precise review.

PR #285, opened 2026-07-31
feat(benchmarks): ILB-Remediation bench + first run (declared vs implemented fix/suggest)

ofri-peretz:
## Summary

- New `ILB-Remediation` suite (measurement D#16/D#18 of the security-benchmark leadership plan): measures remediation coverage — `fixable`/`hasSuggestions` **declared in meta vs actually implemented** in report calls — for the 10 Interlace security plugins and 5 pinned free ESLint-native competitors, same patterns on both sides.
- Registers the bench in `result-schema.json` enum, `state.json`, `bench-context.md`; adds `npm run ilb:remediation`.
- First run committed: `benchmarks/results/ilb-remediation/2026-07-31.json` + history row. Methodology pre-registered (harness committed before the run; `methodologyCommit` pinned).

## Headline result

| | rules | implemented suggestions | implemented fixers | dead declarations |
|---|---|---|---|---|
| Interlace (10 plugins) | 203 | **87** | **7** | 22 (named, remediation tracked) |
| Competitors (5 plugins) | 57 | 0 | 1 | 16 (SDL declares 17/17 fixable, implements 1) |

## Test plan

- [x] `node benchmarks/suites/ilb-remediation/run.mjs --print` — runs clean, deterministic output
- [x] Envelope carries required schema fields (bench/benchVersion/timestamp/toolchain/methodologyCommit); `ilb:validate-results` runs in the quality gate on this PR
- [x] Suite is self-contained (pinned competitor deps in suite package.json + lockfile)

## After merge

Result feeds the D#16 article (publishing queue) and the binary-wins table. Dead-declaration fixes are a separate tracked remediation task; the suite re-runs after it lands to assert declared == implemented.

## Summary by CodeRabbit

* **New Features**
  * Added the ILB-Remediation benchmark to evaluate declared versus implemented fixes and suggestions.
  * Added benchmark reporting for dead and undeclared remediation capabilities across supported tools.
  * Added an npm command to run the remediation benchmark suite.

* **Documentation**
  * Added methodology, measured fields, limitations, execution instructions, and result-format documentation.
  * Added benchmark results, historical records, and updated roadmap and scorecard metadata.


ofri-peretz:
All 8 review threads addressed. Each claim was checked against source before being accepted — one did not survive that check.

**Accepted, and the two that mattered:**

- **Detection scoping.** `RE_IMPL_FIX` / `RE_IMPL_SUGGEST` matched `fix:` / `suggest:` anywhere in a file. Our own devkit is the motivating case: several rules carry a documentation string spelled `fix: 'Use Buffer.alloc(size)…'`, which a file-wide scan reads as a working autofixer. Detection now runs only over `report(...)` argument text via a balanced-paren scan.

  Re-ran under the stricter method: **byte-identical per package** (Interlace 203 rules / 87 implemented suggestions / 7 implemented fixers / 22 dead; competitors 57 / 0 / 1 / 16). Verified independently before changing anything, too — all 7 declared-fixable rules make real `fixer.*` calls, and all 22 dead declarations have no alternative suggestion path. The numbers were right; they now rest on a method that can be defended.

- **Null-source fallback.** `measureCompetitor()` returned `null` when a rule source could not be located. A `null` is excluded by *both* the implemented-count filter and the dead-declaration filter, so an unresolvable rule vanished while the result still looked complete. It now throws with package + rule name. Latent, not manifest: `declared == implemented + dead` held for every package in this run, so nothing was actually dropped.

**Also accepted, each verified rather than taken on faith:** `roadmap.totalItems` 42 → 56 (phases are 21+6+4+6+19); `shippedItems` matched only the exact string `"Shipped"` and missed 9 qualified variants (`startswith` now matches 55); `whenChangingToolchain` still gated on item 1.12, which is Shipped; history rows now carry `resultPath`.

**Rejected — the bench count is 12, not 11.** The review asked to change `.agent/rules/bench-context.md` to 11. `benches.list` in `state.json` holds 12 entries, the `benchmarks/README.md` table holds 11 benches, and ILB-Remediation makes 12. Updated to **12** across `state.json`, `README.md` and `bench-context.md`, repointed the stale `#2-the-10-benches-at-a-glance` anchor, and added the ILB-Remediation row the README table was missing.

**Found while here, which the review and CI both missed:** the `toolchain` block was hand-written as `eslint: null, typescript: null, tsCompiler: 'tsc-classic'`. That is three fatal `validateToolchain()` issues, and the hardcoded value was simply false — this repo is on TypeScript 6.0.3, i.e. **tsc-go**, which is exactly the transition roadmap item 1.14 exists to make visible. Now detected via the shared `getToolchain()`. `npm run ilb:validate-results` went from 2 fatal to **0 fatal across all 79 result files**; both fatals were ours.

The 2026-07-31 result is superseded by 2026-08-02 (correct toolchain, stricter method, same numbers). Neither had reached `main`, so no published history is being rewritten.

PR #306, opened 2026-08-02
feat(docs): branded 404 page for eslint.interlace.tools

ofri-peretz:
## Summary

`eslint.interlace.tools/anything-wrong` served Next.js's stock **"404: This page could not be found"** — an unbranded default page on a public marketing surface. Part of giving all five sites a consistent branded 404.

- Adds `apps/docs/src/app/not-found.tsx` with the two-bar mark, a 404 kicker, copy, and Home / Browse docs links.
- **Imports `InterlaceMark` from the synced `.interlace/` baseline** rather than redefining the SVG, so the mark keeps reading the theme-paired `--brand-mark-bar-*` tokens and there's still one source for the brand geometry.
- Sets `robots: noindex` — a 404 has no content worth ranking, and indexing it competes with the real pages for the same queries.
- Width uses `max-w-prose`, deliberately **not** `max-w-sm..2xl`: the DS spacing tokens shadow those utilities and would silently render a ~96px container.

## Test plan

Verified in the browser against a local dev server:

- [x] Renders correctly at desktop and at 375px — no horizontal overflow.
- [x] Dark mode swaps the mark to the dark-theme token pair automatically (`--brand-mark-bar-o` → `#f4794a`), confirming it inherits the AA-safe pairing rather than hardcoding light-mode hex.
- [x] **Response is still HTTP 404**, not a soft 200.
- [x] `<meta name="robots" content="noindex">` present in the served HTML.

## After merge

Docs deploy publishes the page. No other route changes.


ofri-peretz:
Added `apps/docs/src/__tests__/not-found-lock.test.ts` — good catch, and the right call per CLAUDE.md.

It locks three invariants, each of which would pass typecheck and lint while breaking silently:

- `max-w-prose`, and **none** of `max-w-sm..2xl` (word-boundary matched, so `max-w-xl` can't match inside `max-w-2xl`)
- the mark is imported from the baseline, and the raw geometry `rotate(-30 50 50)` is **absent** — so a re-inlined SVG fails
- `index: false` **and** `follow: true`

Mutation-tested rather than just asserted green: flipping `max-w-prose`→`max-w-2xl` and `follow: true`→`follow: false` each turn the suite red, then restored.

PR #314, opened 2026-08-02
chore(workspace): drop dead `crypto` short from oxlint-parity plugin lists

ofri-peretz:
## Summary

The task that produced this PR was: *"restore `eslint-plugin-crypto` and ship it an `./oxlint` export — it's the only security plugin missing one (10/11)."* **That premise is wrong, and doing it would have been a regression.** This PR fixes the thing that actually caused the misreading.

- **`eslint-plugin-crypto` is not a plugin we ship.** It was deprecated 2026-05-10 and **deleted** on 2026-05-30 (`ddd800d9`, #167); all its rules (weak ciphers/hashes, ECB mode, static IVs, key derivation, timing-unsafe compare, `Math.random()` for crypto, CryptoJS, native-crypto preference) were migrated into `eslint-plugin-node-security`. npm still serves 2.2.3 purely as a deprecated tombstone so existing installs see the migration notice. The "gutted working tree" in the original report was leftover untracked `dist/`/`coverage/`, not an interrupted refactor.
- **The real count is 10/10, and the claim already holds.** Verified end-to-end below — including the crypto rules, which fire under oxlint today from node-security.
- **Root cause of the false gap:** the short name `crypto` survived in two ILB-oxlint-parity plugin lists. The runner filters shorts against the generated manifest (`allowedShorts`), so a stale name is **silently dropped** — no error, no warning. Zero runtime cost, but it reads to humans and agents as "crypto is still a plugin we cover", which is exactly how a retired package ended up proposed for resurrection and republish.

Removed from:
- `benchmarks/suites/ilb-oxlint-parity/run.ts` → `PLUGINS_DEFAULT`
- `package.json` → `ilb:oxlint-parity:full --plugins`

## Test plan

Regression lock (CLAUDE.md contract) — added to `scripts/__tests__/oxlint-export-lock.test.ts`, asserting every short in `PLUGINS_DEFAULT` and in every `--plugins` list across root `package.json` scripts resolves to a live manifest plugin:

- [x] **Fails on the unfixed code** — reintroduced `crypto` into both lists → 2 failures (`run.ts PLUGINS_DEFAULT contains no dead plugin shorts`, `package.json --plugins lists contain no dead plugin shorts`).
- [x] **Passes on the fix** — `npm run test:oxlint-exports` → 61/61.
- [x] Extends the already-CI-wired lock instead of adding a new file: `test:oxlint-exports` names one explicit file, so a new test file would never have run.

Claim verification:

- [x] `npm run oxlint:shims:verify` — 19/19 shims load, rule sets match manifest snapshot.
- [x] `npm run oxlint:verify-runtime` — 33/33 API probes pass.
- [x] **oxlint 1.76.0 (latest), all 10 security plugins loaded together via `jsPlugins`** — `browser-security/no-innerhtml` (CWE-79), `jwt/no-algorithm-none` (CWE-347), and `node-security/no-weak-hash-algorithm` (CWE-327, MD5 — a former crypto-plugin rule) all fire with full CWE/OWASP/CVSS diagnostics.
- [x] `npm run oxlint:shims:check` — no drift.
- [x] `npm run typecheck:scripts` / `npm run typecheck:benchmarks` — clean; confirmed via `--listFiles` that both changed `.ts` files are inside those projects.
- [x] `no-deprecated-plugin-references.test.ts` — 2/2 pass (its markdown + import layers can't see a bare short name in a string list, which is why the new lock is needed).

Pushed with `LEFTHOOK_EXCLUDE=build,portability,typecheck` — all three fail on local worktree environment gaps only (Turbopack vs symlinked `node_modules`; `node_modules/oxlint/dist` and `tsgo` absent in this worktree). CI's `npm ci` provides all three.

## After merge

No package or published surface changes — benchmark config and a test only. `eslint-plugin-crypto` stays deprecated and stays deleted; the answer to any question about it remains "consolidated into `eslint-plugin-node-security`".

Separately (outside this repo), the source doc `adoption-campaign/prepared/oxlint-ecosystem.md` has been corrected: the "crypto gap" finding and its Tier-3 remediation item are retracted, the risk item is struck, and the verified 10/10 result is recorded. *"Every Interlace security plugin runs on oxlint"* is safe to publish as written.


ofri-peretz:
> #### Nit: `liveShorts.size >= 19` floor is currently a pin, not a floor

Good catch — fixed in d20991e1.

You're right that `liveShorts` filters on `ruleCount > 0`, which made `>= 19` against a 19-plugin manifest equivalent to `=== 19`. The failure mode you describe is real: the first plugin scaffolded with no rules yet would have failed this sanity check for the wrong reason and masked the two assertions it's meant to protect.

Lowered the floor to **10**, deliberately well below the current count, with a comment stating why it is *not* a plugin-count pin — its only job is to stop a truncated or empty manifest from making the two dead-short checks pass vacuously. The genuine plugin-count ratchet stays where it belongs: the `pluginDirs.length >= 19` floor at the top of the file, which counts directories and so isn't subject to the `ruleCount` filter.

Also merged `origin/main` (the branch was `BEHIND` after the #289 release). Re-verified post-merge: `test:oxlint-exports` 61/61, `oxlint:shims:check` no drift.

PR #320, opened 2026-08-02
fix(no-cycle): survive deep import chains, and replace the unsupported 100x claim

ofri-peretz:
## Summary

`no-cycle` crashed on deep import chains, and the repo advertised a "100x faster" figure no benchmark supported. This fixes the crash, builds the evidence to say something true instead, and replaces the claim everywhere it appeared.

## The bug

`tarjanStrongConnect` recursed once per graph node. A chain deeper than the JS call stack threw `RangeError: Maximum call stack size exceeded`, and since `no-cycle` defaults to unlimited traversal depth, nothing capped the descent — ESLint exited 2 with **no results at all**. Observed at file 4,974 of a 5,000-node chain (Node 24 / darwin-arm64, ESLint 10.7).

Now runs on an explicit frame stack. Traversal order and every write to `TarjanState` are unchanged, so the SCCs produced are identical; depth is bounded by heap rather than stack.

`eslint-plugin-import` has the same defect in its own `lib/scc.js` and still crashes — that turns out to be the most defensible thing this plugin can claim.

```
$ node benchmarks/scripts/repro-deep-chain.mjs 6000
eslint-plugin-import         FAILED after 93.6s — RangeError: Maximum call stack size exceeded
eslint-plugin-import-next    completed in 18.2s
```

The repro is self-contained — it depends on nothing in this repo.

## Evidence

The synthetic corpus was a single graph shape (deep linear chain), which is the best case for a plugin that amortizes traversal across files. Added four more at the same 5,000-file count. Medians, n=3 unless noted, **detection parity verified per shape before any timing was trusted**:

| Shape | `eslint-plugin-import` | `import-next` | |
|---|---|---|---|
| deep chain | **crash** | 11.34s | we finish, it can't |
| wide/shallow | 2.81s | 2.80s | tie (n=7, interleaved) |
| flat, no imports | 2.28s | 2.23s | tie |
| dense, 1,000 cycles | 3.96s | 3.64s | 1.1× |
| cold single file | 0.39s | 0.36s | 1.1× |

Never slower on any shape. Two ties, two modest wins, one graph the official plugin cannot complete. Variance is also ~12× tighter on the wide shape (stdev 0.11 vs 1.37).

## Harness defects fixed

Three, all of which produced confidently wrong numbers:

- **Crashes were timed as fast runs.** `runEslint` swallowed every non-zero exit as "expected lint errors" — true for exit 1, but it also caught exit 2 (rule threw). A plugin that crashes exits *early*, so the crash scored as a short duration. The bias only ever ran one way.
- **stdout `maxBuffer` overflow is indistinguishable from a timeout.** 20,000 report lines overrun `execSync`'s 1 MB cap; Node kills the child with SIGTERM. `dense-5000` read as ">300s" for both plugins when both finish in under 4s.
- **Fixtures were generated where the runner does not look.** The generator wrote to `benchmarks/<name>/fixtures` (a path that has never existed) while the runner read `suites/<suite>/fixtures`. The synthetic suite could not be regenerated from committed source at all.

Methodology now also records that medians must be used over means: on `wide-5000` a single 6.05s outlier dragged the official plugin's mean to 3.70s against a 2.81s median and inverted the verdict.

## Claim cleanup

"100x faster" appeared in **16 places** and nothing supported it. Highest figure ever measured is 54.9× — isolated rule, synthetic best-case shape, 5,000 files. The 10,000-file run was terminated before completing, which is the likely origin of the extrapolation.

Two rows in `EMPLOYER_SIGNAL.md` were worse than an extrapolation, and that file scripts interview talking points:

- `"45s+ on large monorepos" → "<1s (100x faster)"` — no benchmark shows sub-second anything.
- `"Memory: OOM on large repos | Bounded"` — peak RSS is 4,035 MB official vs 4,064 MB ours. We use **29 MB more**. Not unsupported, backwards. Now states parity explicitly.

It also listed "100x Performance Claims — quantifiable, benchmarkable value" as a differentiator, which was the one claim in it that could not be benchmarked.

The Dev.to article slug still contains `up-to-100x-faster` and is **deliberately left alone** — changing it breaks the URL, and the article's title and body already use the correct 54.8× figure.

## Test notes

- 10,000-node regression test reduced to 6,000 — **verified to still reproduce the `RangeError` against the pre-fix implementation**, so the guard is real rather than decorative. At 10,000 its I/O starved neighbouring tests past their timeouts under `turbo run test`.
- `no-deprecated-plugin-references` walks the whole repo tree with vitest's 5s default. A sibling doing heavy I/O in a parallel worker could starve it and fail an unrelated commit. Given an explicit 60s budget.

## Verification

- `@interlace/eslint-devkit` — 934 passed, 100% coverage (lines/statements/branches/functions)
- `eslint-plugin-import-next` — 1,300 passed
- Detection parity exact on every shape that completes

## Known limits

The matrix is five synthetic shapes plus one real 455K-line codebase. "Never slower on any graph shape tested" is exactly as strong as that corpus and no stronger. The real-codebase corpus is a private repo; the committed JSON is the public artifact.

Out of scope, tracked separately: seven `src/__compatibility__/*.spec.ts` suites time out on cold module caches and block unrelated commits.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `no-cycle` handling for very deep dependency chains, preventing call-stack failures while preserving cycle detection behavior.
  * Added support for negative depth-limit handling.

* **Documentation**
  * Documented deep dependency graph support, reproduction guidance, benchmark methodology, and performance results.

* **Tests**
  * Added coverage for large acyclic dependency chains and extended regression-test timeouts.

* **Chores**
  * Added benchmark scenarios for multiple dependency-graph shapes and improved reporting for failed comparisons.

### DIFF
--- a/.changeset/no-cycle-iterative-scc.md
+++ b/.changeset/no-cycle-iterative-scc.md
@@ -1,0 +1,10 @@
+---
+'@interlace/eslint-devkit': patch
+'eslint-plugin-import-next': patch
+---
+
+**`no-cycle` no longer crashes on deep import chains.** Tarjan's SCC pass recursed once per graph node, so a chain deeper than the JS call stack threw `RangeError: Maximum call stack size exceeded` — and since the rule defaults to unlimited traversal depth, nothing capped the descent. ESLint exited 2 with no results at all: not a slow lint, no lint. Observed at file 4,974 of a 5,000-node chain on Node 24.
+
+The traversal now runs on an explicit frame stack. Traversal order and every write to the Tarjan state are unchanged, so the components produced are identical; depth is bounded by heap rather than by the call stack. `eslint-plugin-import` has the same defect in its own `lib/scc.js` and still crashes on the same input.
+
+Chains reach these depths through generated API clients, nested barrel files, and long `export … from` ladders — depth accumulates through re-export edges, which is exactly what the rule follows. Reproduce with `node benchmarks/scripts/repro-deep-chain.mjs 6000`.
--- a/.gitignore
+++ b/.gitignore
@@ -156,3 +156,6 @@ packages/*/src/**/*.js
 packages/*/src/**/*.js.map
 packages/*/src/**/*.d.ts
 .review-screens/
+
+# scratch dir from benchmarks/scripts/repro-deep-chain.mjs (self-cleans, but not if interrupted)
+.deep-chain-repro/
--- a/CLAIMS.md
+++ b/CLAIMS.md
@@ -59,16 +59,31 @@ early, so timing it would score the crash as a win.
 
 | Claim (as it appears in docs/marketing) | Suite | Latest result | Last verified |
 | --- | --- | --- | --- |
-| "`eslint-plugin-import` crashes on deep import chains; `import-next` completes them" | ilb-perf-import-shapes (`chain-5000`) | [2026-08-02.json](benchmarks/results/ilb-perf-import-shapes/2026-08-02.json) — official `failed: exit 2` (`RangeError` in its own `lib/scc.js`), ours 11.34s. Standalone repro: [`benchmarks/scripts/repro-deep-chain.mjs`](benchmarks/scripts/repro-deep-chain.mjs) | 2026-08-02 |
-| "Never slower than `eslint-plugin-import` on any graph shape tested" | ilb-perf-import-shapes (5 shapes @ 5,000 files) | [2026-08-02.json](benchmarks/results/ilb-perf-import-shapes/2026-08-02.json) — chain: official crashes · wide 2.81s vs 2.80s · flat 2.28s vs 2.23s · dense 3.96s vs 3.64s · single 0.39s vs 0.36s | 2026-08-02 |
-| "~1.1x faster on dense cyclic graphs and cold single-file runs" | ilb-perf-import-shapes (`dense-5000`, `single`) | [2026-08-02.json](benchmarks/results/ilb-perf-import-shapes/2026-08-02.json) | 2026-08-02 |
-| "Run-to-run variance ~12x tighter" | ilb-perf-import-shapes (`wide-5000`, n=7 interleaved) | stdev 0.11 ours vs 1.37 official; official threw a 6.05s outlier against a 2.81s median | 2026-08-02 |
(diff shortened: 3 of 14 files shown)


ofri-peretz:
All review findings addressed in `4de44680`. Summary of what changed and what did not:

**Fixed**
- **Absolute paths in the committed result** — a username and private worktree path were embedded in a public artifact. Redacted; only the `eslint-plugin-import/lib/scc.js` stack location remains. Good catch, this one mattered most.
- **Missing `schemaVersion` / `toolchain`** — added. This was the first dated result to omit fields `CLAIMS.md` requires.
- **`repro-deep-chain.mjs` accepted garbage input** — `Number('abc')` is `NaN`, so the loop never ran and *both plugins reported success*. A repro that passes on input it never tested is worse than no repro. Now exits 2.
- **Failed plugins dropped from the summary table** — the crash is this suite's headline result and the table silently omitted it. Now prints `FAILED`, and medians rather than means.
- **Means vs medians** — fixed in `0a3d2a0a`, along with interleaving the plugins per round. The sequential loop handed the second slot to the same plugin every time, which is a systematic bias rather than noise.
- **`EMPLOYER_SIGNAL`** — removed the surviving "45s → 0.4s" tactic line, and scoped the deep-graph row to `no-cycle` rather than implying every rule in the package.
- **Parity gate** — the methodology claimed it was enforced; the runner records times and failures but not violation counts, so no result file proves it. Now documented as a manual, operator-owned check that must be redone whenever fixtures are regenerated.

**Retracted rather than defended**
`wide-5000` no longer carries any claim. Recomputed with medians the committed run says **0.8× — us slower**; an interleaved n=7 run says **1.01×, a tie**. They disagree, and the re-run that would settle it was attempted and discarded: it recorded 46s and 78s on a shape that measures ~2.8s idle, under a load average of 340 from concurrent builds. `CLAIMS.md` now marks the shape unresolved and explicitly forbids "never slower on any graph shape tested" — the only committed number for that shape contradicts it.

**Acknowledged, not changed**
The `relax(frame.file, 0)` depth-limit fallback. Your analysis is right: under a finite `maxDepth` the SCCs are not guaranteed correct, and `Number.MAX_SAFE_INTEGER` is the proper neutral element. I preserved the pre-existing behaviour deliberately so this PR's SCC output stays byte-identical to the recursive version — that property is what makes the crash fix safe to merge. Changing it needs its own PR and its own test, and your comment is the right place for that to start.

PR #322, opened 2026-08-02
feat(sequelize-security): close the CWE-89 gap Juice Shop exposed

ofri-peretz:
## The finding

Scanning OWASP Juice Shop with the recommended presets of `secure-coding`, `node-security`, `express-security` and `mongodb-security` produced **zero** findings on its two flagship SQL injections:

- `routes/search.ts:23` — `models.sequelize.query(\`SELECT * FROM Products WHERE ((name LIKE '%${criteria}%' ...\`)`
- `routes/login.ts:34` — `models.sequelize.query(\`SELECT * FROM Users WHERE email = '${req.body.email}' ...\`)`

A probe run of `eslint-plugin-pg` flagged **both** correctly.

## The diagnosis

Detection was never the problem. `pg/no-unsafe-query` matches any `.query()` member call and there is nothing Postgres-specific in its AST work — it catches `sequelize.query()` today. The problem is **distribution**: nobody on Sequelize, SQLite or MySQL installs the Postgres plugin.

## The fix — two commits

**1. `refactor(devkit,pg)`** — extract the detector into `@interlace/eslint-devkit` as `createSqlInjectionRule`, parameterised on the three things that actually differ between drivers: the sink list, a SQL-keyword precision gate, and the remediation copy. `pg/no-unsafe-query` becomes an instantiation of it — same rule id, message ids, sink and behaviour. **All 28 pre-existing pg rule tests pass untouched**, which is the regression proof for the refactor.

**2. `feat(sequelize-security)`** — new plugin, `sequelize-security/no-unsafe-query` (CWE-89, `error` in `recommended`):

- **Sinks:** `sequelize.query()` and `Sequelize.literal()` — the latter is where `ORDER BY` / column-name injection lives, so no keyword gate is applied (an interpolated column name carries no SQL keyword of its own).
- **Remediation:** Sequelize's own conventions (`replacements`, `bind`), not generic "use parameterized queries" advice.
- **Taint tracking:** queries assembled across statements, including with `+=`.

Both Juice Shop sites are pinned verbatim as test cases.

## Placement

The rule keys on **driver method names, not language primitives**, so it belongs in driver-scoped plugins — not in `secure-coding` / `node-security` / `browser-security`. An earlier draft of this branch put a generic `secure-coding/no-sql-injection` in place; that was wrong under the taxonomy rule and has been dropped. It also produced double-reporting on pg codebases, which was a symptom of the misplacement rather than something to work around.

For the same reason the new plugin is **not** added to `eslint-config-interlace`'s aggregated presets: its `.query()` sink overlaps pg's, so bundling it would report the same line twice. Install it when you use Sequelize.

## Naming

`eslint-plugin-sequelize` is **taken on npm** (owner `tebriel`, v1.0.4), hence the `-security` suffix — which also matches the `express-security` / `mongodb-security` / `nestjs-security` / `lambda-security` family. `sequelize` is declared in `peerDependencies` with a supported major range **and** marked optional in `peerDependenciesMeta` (a `Meta` entry alone is silently ignored by npm).

## Also in here

- Raises the timeout on devkit's `no-deprecated-plugin-references` guard. Both its layers shell out to a repo-wide `grep`, which cannot finish inside vitest's 5s default once the suite has enough files running in parallel; this branch's new test file tipped it over, and it failed as a *timeout*, not a violation.

## Verification

- **49/49** turbo test tasks green; new plugin at 100% coverage (25 tests); devkit and pg still 100%.
- 3/3 vulnerable and 0/3 safe fixtures in `benchmarks/corpus/CWE-089` — whose manifest already listed the gap before it was closed.
- All 10 pre-push gates green: build, typecheck, tests, lockfile, portability, workflows, changelog, markdown, shim-drift, shim-verify.

Found during the adoption-campaign vulnerable-apps scan (FN-1 / action A1).


ofri-peretz:
All three findings addressed in 6123756c:

**1. Numeric `+` classified as concat (real FP)** — confirmed and fixed. `classify` now requires at least one string literal in the expression before treating `+` as SQL concatenation; template literals stay exempt since a template is always a string. `db.query(1 + offset)` no longer reports.

Trade-off worth naming: `db.query(a + b)` (two identifiers, no static text) is also no longer reported. Nothing in that expression indicates SQL, and the FP cost outweighed a detection that only fired when the entire query came from variables. Pinned by two new valid cases on the *ungated* instance, which is where the FP actually bit.

**2. Missing taint → `literal` test** — added. Taint propagation through `Sequelize.literal()` (the ORDER BY vector) is now locked, not just `query`.

**3. v6 docs link vs `^7.0.0` peer** — fixed to the v7 URL in both the rule and `docs/rules/no-unsafe-query.md`.

61/61 turbo test tasks green; devkit suite at 30 tests.

PR #346, opened 2026-08-04
perf(eslint-devkit): seed the deep-chain guard in memory, 34.3s → 1.7s

ofri-peretz:
## Summary

While profiling the CI pipeline I traced the `Unit Tests (Turbo)` bottleneck (551s, the entire merge-blocking critical path) down to one test file, then to one test inside it.

`dependency-analysis.coverage.test.ts` was **34.3s of devkit's 33.8s wall**, and devkit is **27% of the whole `turbo run test` cost**. Inside it, the deep-import-chain stack guard was **25.2s of 25.5s**, because it materialized a 6,000-file import chain on disk.

Measured at `DEPTH = 6_000` on an idle machine:

| phase | cost |
|---|---:|
| write 6,000 files | 2,437ms |
| read 6,000 files (331KB) | 9,572ms |
| `rmSync` 6,000 files — the shared `afterEach` | 3,700ms |

It's per-file syscall overhead, not volume — 331KB total.

That teardown is also the source of the intermittent `Hook timed out in 10000ms` that has been blocking unrelated pushes: vitest's `hookTimeout` defaults to 10s and **0 of 28** vitest configs in this repo raise it. Under the 58-task fan-out the same test was observed at **86,293ms**.

## The fix

None of that I/O proved anything about the guard. The guard is that Tarjan doesn't recurse once per node. `getFileImports()` returns `cache.dependencies.get(file)` *before* it touches the filesystem, and the SCC walk follows `ImportInfo.path` — so seeding the chain straight into the resolver cache gives Tarjan the identical 6,000-node traversal with zero files.

No production code changed. No API change. The changeset is empty by design (test-only, nothing published shifts).

## Test plan

- [x] File: **34.3s → 1.7s**. Package: devkit **36.0s → 25.3s** wall.
- [x] `1017 passed | 2 skipped` — identical to before.
- [x] Full devkit suite green after a clean `npm run build --workspace=@interlace/eslint-devkit`.
- [x] `turbo run test --filter="...[origin/main]"` — 61/61 successful.

### The guard still has teeth — verified, not assumed

The risk in this change is making the test fast by making it prove nothing. So I ran a **recursive** Tarjan (the implementation this guard exists to reject) over the same in-memory 6,000-node chain:

```
RangeError: Maximum call stack size exceeded
  -> died after indexing 4130 of 6000 nodes
```

`DEPTH = 6,000` keeps a ~45% margin over the recursion ceiling, consistent with the 4,974-of-5,000 figure the original comment recorded. The number stays tied to that rationale.

## Why this matters for the CI work

devkit's cost was so concentrated in this one file that **`vitest --shard` did nothing for it**:

```
before fix:  shard 1/4 = 34.6s   2/4 = 2.0s   3/4 = 1.8s   4/4 = 1.9s
after fix:   shard 1/4 = 10.0s   2/4 = 8.3s   3/4 = 4.0s   4/4 = 4.3s
```

Shards split by *file*; one file carried 95% of the cost, so no shard count divided it. Package-level shard packing flatlined the same way — 4, 6, 8 and 12 shards all bottomed out at 103.8s, devkit's indivisible task.

**Fix distribution first, then shard.** This PR is that first step; sharding is now worth adopting.

### DIFF
--- a/.changeset/devkit-deep-chain-fixture.md
+++ b/.changeset/devkit-deep-chain-fixture.md
@@ -1,0 +1,6 @@
+---
+---
+
+Test-only: the devkit deep-import-chain regression guard now seeds the resolver
+cache in memory instead of materializing 6,000 files on disk. No runtime or API
+change.
--- a/packages/eslint-devkit/src/resolver/dependency-analysis.coverage.test.ts
+++ b/packages/eslint-devkit/src/resolver/dependency-analysis.coverage.test.ts
@@ -231,28 +231,37 @@ describe('computeSCCsFromFile edge cases', () => {
       // no-cycle defaults to an unlimited maxDepth, so nothing capped it — the
       // whole lint run died.
       //
-      // 6,000 clears that observed ceiling by ~20% while costing 6,000 file
-      // writes and reads rather than 10,000. That matters: this suite runs
-      // under `turbo run test`, which executes every package's tests in
-      // parallel, and at 10,000 nodes the I/O contention pushed a ~2s test past
-      // a 120s timeout. If a future platform raises the stack limit past 6,000
-      // this guard weakens, so treat the number as tied to the figure above.
+      // 6,000 clears that observed ceiling by ~20%. If a future platform raises
+      // the stack limit past 6,000 this guard weakens, so treat the number as
+      // tied to the figure above.
       const DEPTH = 6_000;
 
-      // Written directly rather than through createTempFile: that helper calls
-      // realpathSync per file, and 10,000 extra syscalls slow this test enough
-      // to push unrelated tests in the suite past their own timeouts.
+      // The chain is seeded straight into the resolver cache instead of being
+      // materialized on disk. getFileImports() returns cache.dependencies.get()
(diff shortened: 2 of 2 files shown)


ofri-peretz:
Superseded by #351, which landed the identical fix — seeding the 6,000-node chain into `cache.dependencies` instead of materializing files — while this sat in review. Closing to avoid a conflicting edit to the same file.

Both arrived at the same design independently, which is a decent signal it's the right one. #351 goes further: it seeds the reads for both `computeSCCsFromFile`'s discovery pass *and* `tarjanStrongConnect`'s edge reads, so its fixture touches zero files (86293ms → 39ms) where mine still left the file at 1.7s.

One thing worth carrying over from here: I verified the guard retains its teeth by running a recursive Tarjan over the same seeded chain — it dies with `RangeError: Maximum call stack size exceeded` after indexing 4,130 of 6,000 nodes on Node 24 / darwin-arm64. #351 records 3,104. Either way DEPTH=6,000 keeps a comfortable margin over the recursion ceiling, so the number should stay tied to that rationale rather than being trimmed for speed later.

PR #353, opened 2026-08-04
feat(workspace): no-unscoped-mutation for prisma, drizzle and knex

ofri-peretz:
## Summary

Every ORM ships a bulk mutation whose *unscoped* form rewrites or deletes the whole table — `prisma.user.deleteMany()`, `db.delete(users)`, `knex('users').del()`. Each type-checks, passes review, and only shows up once it has run against production data. `eslint-plugin-drizzle`'s entire published surface is this one check for one ORM; this generalizes it.

Detection lives in one place — `createUnscopedMutationRule` in `@interlace/eslint-devkit` — and each plugin instantiates it with its own sinks and remediation copy, the shape `createSqlInjectionRule` already established.

| Plugin | Sinks | Where scope comes from |
| :--- | :--- | :--- |
| `prisma-security` | `deleteMany`, `updateMany` | `{ where }` in the options object |
| `drizzle-security` | `delete`, `update` | a chained `.where()` |
| `knex-security` | `del`, `delete`, `update` | any of the chained `where*` family |

`argumentRole` is the one thing that cannot be inferred: a lone identifier argument is the **filter** for Prisma (`deleteMany(opts)`) and the **table** for Drizzle (`db.delete(users)`). Reading it wrong either suppresses the headline Drizzle finding or invents a false positive on every dynamically built filter.

## Not shipped for Sequelize or TypeORM

Sequelize gives its instance and static forms the same names, and both accept an options object — `user.destroy({ transaction: t })` (one row) and `User.destroy({})` (the whole table) are the same AST. Two false positives surfaced in its suite and the rule was **withdrawn from that package rather than shipped with them**: a rule that fires on correct code is the one users disable, which is the exact wedge we compete on. The genuinely detectable case, `destroy({ truncate: true })`, becomes its own rule. TypeORM's bare-criteria shape is a third detection shape, deferred for the same reason.

Both are recorded in `ORM-AGENTIC-RULE-PLAN.md` rather than left as silent gaps. Scope that cannot be read statically is treated as present, so the rule stays silent rather than guessing.

## Test plan

- [x] devkit 1057, prisma 29, drizzle 26, knex 29 — all at 100/100/100/100
- [x] Self-suppression lock verified by reverting the chain-ascent guard: exactly the one lock test fails, the other 36 stay green
- [x] docs 1650 passed, including the rule-source-drift and README generator locks
- [x] `turbo run test` over all affected packages: 61/61 green
- [x] Full pre-push gate green: typecheck, portability, changelog, shim-verify, shim-drift, tests, lockfile, build
- [x] READMEs regenerated by `sync-readmes`, not hand-edited

## After merge

Ships in `strict` only. Promotion to `recommended`/`flagship` waits on a measured false-positive profile against the benchmark corpus — that is W6 in the plan.

One catch worth noting for other in-flight work: `no-runtime-optional-peer` (from #341) landed on `main` while this was in progress and caught a real defect here — the factory was importing `AST_NODE_TYPES` as a **value** from `@typescript-eslint/utils`, an optional peer npm never installs, which would have failed to load in every consuming plugin. Now imported from the local `ast-node-types` shim. That lock earned its keep.

## Summary by CodeRabbit

* **New Features**
  * Added the `no-unscoped-mutation` security rule for Prisma, Drizzle, and Knex.
  * Flags bulk deletes and updates that lack a static scope filter, such as `where` clauses.
  * Enabled the rule in strict configurations.

* **Documentation**
  * Added usage guidance, examples, security references, and known limitations for each supported framework.

* **Tests**
  * Added comprehensive coverage for scoped, unscoped, and non-mutation operations.


ofri-peretz:
Thanks — three of these were real bugs, not nits. Fixed in `b5ecd895`.

**Generic sink names matched any receiver** (`no-unscoped-mutation.test.ts`, drizzle/knex)

The most serious one. `delete` and `update` are also `Map.prototype.delete`, `Set.prototype.delete`, and the vocabulary of every store wrapper, so `map.delete(key)` and `store.update(patch)` reported as unscoped mutations in ordinary JavaScript. `mongodb-security/src/utils/receiver.ts` already documents exactly this doctrine — *"method names alone are hopeless discriminators"* — and I didn't apply it.

The factory now takes `modules` and `receiverPattern`. It skips files that never import the driver, and skips receivers whose chain base doesn't read as a driver handle. Regressions added per plugin: `map.delete`, `set.delete`, `store.update`, a non-driver receiver, and a file with no driver import at all.

**An empty `where` was treated as scope** (prisma)

Correct, and it defeated the rule's whole purpose — `deleteMany({ where: {} })` deletes every row. Key presence is no longer sufficient; an empty object literal is not a filter. Added as an `invalid` case in both the plugin and the factory.

**Knex's `where` family was incomplete** (knex)

Also correct. `orWhereIn`, `orWhereNull`, `orWhereRaw`, `whereWrapped`, `whereILike` and eleven more were missing, so correctly scoped queries were reported. `scopeMethods` now mirrors the builder's full family (35 entries), with the five you named covered by explicit `valid` fixtures.

**`hasTruthyKey` and falsy non-`false` values**

Both reviewers flagged this. Fixed to `!prop.value.value`, so `truncate: null` and `truncate: 0` are disabled flags. You were right that it has no shipped impact yet — `truncateKeys` is unused until Sequelize's `no-truncate` lands in W4 — but the fixture is in place now.

**Docs**

All three taken: the changeset described a scope shape neither plugin uses; the plan's O2 row still claimed seven drivers when three shipped; the new options types had no TSDoc.

Verification: factory 49 tests at 100/100/100/100 (including the import gate, CommonJS `require` forms, `this`-receivers, unresolvable bases); prisma 31, drizzle 30, knex 35; full workspace suite 61/61; every pre-push gate green.

Not taken: `benchmark-results/plugin-scope-audit.json` embeds a `generatedAt` timestamp, which makes it churn on every hook run and reliably conflict with lefthook's stash restore. Real problem, but it predates this branch and belongs in its own PR.

PR #385, opened 2026-08-05
feat(workspace): no-raw-identifier-interpolation for drizzle and prisma

ofri-peretz:
## Summary

W1 / O1 from [ORM-AGENTIC-RULE-PLAN.md](ORM-AGENTIC-RULE-PLAN.md).

A bind parameter can only ever substitute a **value**. `$1` is a placeholder in the value slot of the parse tree, so no database accepts one where a table, a column, or a sort direction belongs — and when the hole is an identifier, the driver has nothing to bind and splices the string in verbatim:

```ts
await prisma.$queryRaw`SELECT * FROM users WHERE id = ${id}`;  // parameterized
await prisma.$queryRaw`SELECT * FROM ${table}`;                // injectable
```

Both lines use the API the docs call safe, and both survive review. This is the shape behind Drizzle's [GHSA-gpj5-g38j-94v9](https://github.com/advisories/GHSA-gpj5-g38j-94v9), and it is invisible to every SQL-injection linter that decides by asking *"is this a raw API"* — this **is** the safe API.

## Why it does not overlap `no-unsafe-query`

`no-unsafe-query`'s sinks are the raw string entry points (`$queryRawUnsafe`, `sql.raw()`), where every interpolation already reports. O1 gates on the *tagged template*, where nothing fires today. Zero shared lines, so the taxonomy contract holds.

Silent by design: value holes, string literals, nested `sql` fragments, and Drizzle's composition surface (`sql.identifier`, `sql.join`, `sql.fromList`, `sql.placeholder`). `sql.raw()` deliberately is **not** exempt.

## Scope: 2 drivers, not the 7 the wave estimated

Only Prisma and Drizzle ship a value-parameterizing tagged template. For the other five the raw entry point takes a plain string, where `no-unsafe-query` already reports every interpolation — adding O1 there buys no detection and double-reports one line from one plugin. Recorded as plan §5.2, along with the real residual gap: for those five the remediation reads "use a bind parameter", which is *impossible advice* in an identifier position. That is a message defect in an existing rule, tracked for W4.

## Test plan

- [x] `createRawIdentifierRule` at **100/100/100/100** in `@interlace/eslint-devkit` (35 factory tests: helper-level + both tag-gate shapes through RuleTester).
- [x] Drizzle 57 tests, Prisma suites green; barrel locks updated to 3 rules each and re-verified.
- [x] Full `turbo run build test` — 59/59 tasks.
- [x] All 8 pre-push gates green (`typecheck`, `build-and-test`, `lockfile`, `shim-drift`, `portability`, `changelog`, `markdown-full`, `workflows`).
- [x] Generated artifacts regenerated to a fixpoint: `oxlint:shims` → `oxlint:shims:verify --update` (470 → 472 rules), `sync-plugin-stats` (501 → 503), `sync-readmes`, scope audit.

## After merge

New rules land in `strict` only. Promotion to `recommended`/`flagship` waits on W6's measured FP profile, per plan §1.6.

Also corrects a stale note in `.agent/plugin-rule-manifest.json`: mysql2 has no top-level `rejectUnauthorized` (the docs were fixed for this in #373).

## Summary by CodeRabbit

* **New Features**
  * Added a new security rule for Drizzle and Prisma that flags when SQL identifiers (like table/column names and sort directions) are interpolated in tagged template queries.
  * Released minor updates that include the new rule in both plugins and refreshed the published rule listings and rule statistics.

* **Documentation**
  * Expanded documentation for both plugins with rationale, examples, and recommended safer alternatives.

* **Bug Fixes**
  * Improved detection coverage for identifiers and sort-direction interpolations while avoiding false positives for safe value parameters and approved patterns.


ofri-peretz:
Both gate findings were correct and share one root cause — fixed in `138b83f0`.

The import gate keyed on **AST shape** (`node.tag.type === Identifier`) when it should have keyed on the **tag name**. `q.sql` is a MemberExpression, so it matched `tags` and then skipped the gate entirely — reporting under `drizzle-security` in a file with no Drizzle in it. Replaced with an explicit `requireImport` flag: drizzle sets it, prisma does not. That also makes the second point explicit rather than implied — `modules` genuinely was a no-op for Prisma, and the flag now says so at the call site.

Locks added in both the factory and the drizzle suite, and verified by reverting: the drizzle case fails on the old gate, passes on the new one.

On the third comment — `DISTINCT ON` is already covered. `'SELECT DISTINCT ON '` is in the `identifierPosition` matrix in `raw-identifier-rule.test.ts`, inside the same loop the comment quotes.

ofri-peretz:
All five addressed in `25805e24`.

The `sql.raw()` one was a genuine contradiction and you were right not to guess which side I meant. The **detector** was correct; the **docs** were wrong about why. The two rules divide by what is wrong, not by which API was used:

| | reports |
|---|---|
| `no-unsafe-query` | **string construction** — concatenation/interpolation used to *build* the text passed to `sql.raw(...)` |
| this rule | **position** — a hole where an identifier belongs |

`${sql.raw(table)}` in a FROM slot is a finding here; `sql.raw('SELECT * FROM ' + table)` is a finding there. A line doing both is two defects with two fixes, not one finding twice. That's now a table in the docs, and the contradictory bullet is deleted.

The Prisma example was the other real one: I labelled an allowlisted `$queryRawUnsafe` call as clean, and it isn't — `no-unsafe-query` reports every interpolation reaching that sink and can't see the allowlist. The ✅ block now uses the query builder, and the raw variant moved to its own section carrying the `eslint-disable-next-line` it actually needs, with the reason next to the code.

Malformed code span and the duplicate §5.2 both fixed (installation section is now §5.3).

On the drizzle catalog version: `plugin-stats.json` now reads `0.2.2`, matching `src/index.ts` and `package.json`. It was stale because main's release landed mid-branch; regenerating after the merge resolved it.

ofri-peretz:
Fixed in `e52dfd18` — you're right that the count check proves nothing on its own. It passes if strict drops one rule and picks up any other `drizzle-security/*` key, which is precisely how a new rule gets silently left out of the preset it's meant to join.

Replaced the two spot-checks with an iteration over every key in `rules`, so membership and count are asserted by the same loop. Applied to the prisma suite too, which had the identical weakness.

PR #386, opened 2026-08-05
feat(workspace): no-hardcoded-credentials for knex, mysql2, sequelize and typeorm

ofri-peretz:
## Summary

W1 / O5 from [ORM-AGENTIC-RULE-PLAN.md](ORM-AGENTIC-RULE-PLAN.md). New shared factory `createHardcodedCredentialsRule`, instantiated for four drivers.

A password in source is a password in git history, in every fork, and in every layer of the built image. It's unusual among findings in that fixing it later doesn't help — deleting the line leaves the secret one `git log -p` away for anyone who's ever had read access, so a real fix means rotating the credential **and** rewriting history. The only cheap moment is before it lands.

## Two false positives tightened vs. the pg original

This generalizes the detection `eslint-plugin-postgresql-security` already ships, and fixes two shapes it reports that aren't findings:

| Shape | pg today | Here |
|---|---|---|
| `postgres://localhost:5432/app` | reported | silent — no credential in it |
| `postgres://app@db/app` | reported | silent — username, no password |
| `password: ''` | reported | silent — local trust-auth sentinel |
| `postgres://app:s3cret@db/app` | reported | reported |

## The credential can't be its own evidence

`password`, `user` and `username` are excluded from the connection-shape hints, so an object has to name somewhere to connect **to** — `host`, `port`, `database`, `connectionString` — before its password counts.

Without that, `{ user, password }` makes the login form of every app that also has a database a finding. My own test caught this: `{ password, confirm }` reported until the hint list was narrowed. That case is now a permanent `valid` fixture, along with a login form and a test-fixture user record.

## Test plan

- [x] Factory at **100/100/100/100** (37 tests). Branch coverage closed by exercising the `connectionKeys` default, a spread member, and a computed key through the rule rather than only the helpers.
- [x] `turbo run test` across devkit + the 4 plugins: 5/5.
- [x] Barrel locks updated (knex → 4 rules, others → 3) and re-verified.
- [x] All pre-commit and pre-push gates green.
- [x] Regenerated to a fixpoint: `oxlint:shims` → `oxlint:shims:verify --update` (472 → 474), `sync-plugin-stats` (503 → 505), `sync-readmes`, scope audit.

## Follow-up, deliberately not in this PR

Migrating `postgresql-security/no-hardcoded-credentials` onto the factory is what actually removes its two false positives for pg users. It changes the behaviour of a published rule, so it wants its own PR and its own changeset.

## Summary by CodeRabbit

* **New Features**
  * Added a `no-hardcoded-credentials` security rule for Knex, MySQL, Sequelize, and TypeORM.
  * Detects literal database passwords in connection settings and credential-bearing URLs.
  * Ignores empty passwords, runtime values, credential-free URLs, and unrelated objects.

* **Documentation**
  * Added usage guidance, examples, severity details, exclusions, and remediation recommendations for each plugin.

* **Tests**
  * Added comprehensive coverage for credential detection and safe configuration patterns.


ofri-peretz:
All addressed in `ec9c1890`. Three were behaviour changes, not polish:

**Static template literals** — you were right that this was a full bypass, not an edge case. `password: \`hunter2\`` and `connection: \`postgres://app:s3cret@db/app\`` skipped *both* CWE-798 paths; one character turned the rule off. Added `staticStringValue`, which accepts a template with zero interpolations and still rejects an interpolated one.

**`extra: { webhook: { secret } }`** — correct, and it was the worse half of the same design. Recursion now follows connection-named keys only, so `{ connection: { password } }` still reports and a third-party SDK secret nested under a driver key does not.

**O(N²) rescanning** — fixed with a `scanned` WeakSet reset in `Program`. `reported` was only suppressing the duplicate diagnostic, not the duplicate work.

**`client` as a knex hint** — removed, same reasoning that already excluded `password`/`user`/`username`: a real knex config always names `connection` too, and `client` is the most common name in JS for an HTTP client. Your suggested fixture is in as a `valid` case.

**devkit range `^1.6.1`** — real, and it would have shipped broken. Bumped to `^1.7.0` on mysql/sequelize/typeorm.

**mysql docs using `postgres://`** — corrected to `mysql://` and the "not reported" text now says the scoping out loud.

**`host:5432@other/db`** — documented as a known, deliberate over-match in the `urlEmbedsCredentials` JSDoc rather than parsed around. Also added a lock for the empty-username form `postgres://:pw@host/db`, which is a real password and does report.

Every one of these has a lock. Factory still 100/100/100/100.

PR #389, opened 2026-08-05
feat(workspace): no-mass-assignment for the five ORM plugins with object writes

ofri-peretz:
## Summary

W1 / O3 from [ORM-AGENTIC-RULE-PLAN.md](ORM-AGENTIC-RULE-PLAN.md) — the last shared ORM rule. New factory `createMassAssignmentRule`, instantiated for prisma, drizzle, sequelize, typeorm and knex.

```ts
await prisma.user.update({ where: { id }, data: req.body });
await User.create(req.body);
await db.insert(users).values({ ...req.body });
```

Each line updates the fields the endpoint is *about* — and every other column on the model: `role`, `isAdmin`, `ownerId`, `emailVerified`, `credits`, `stripeCustomerId`. None appear in the diff, which is why the shape survives review. The vulnerability is in what the code does not say.

It also gets worse with nobody touching it: add a `role` column six months from now and every existing mass-assignment site silently starts accepting it. No line changes, and the exposure is new. That's what makes it worth a rule rather than a review habit.

## Silent by design

| Shape | Why |
|---|---|
| `{ name: req.body.name }` | the fix — fields are named |
| `{ ...req.body, updatedAt }` | **still reported** — a named field beside a spread doesn't narrow it |
| `form.body`, `config.query` | ordinary objects; the chain must bottom out in a request identifier |
| `ctx.data`, `context.data` | `data` is app state in several frameworks, so it isn't a request surface |
| `repo.create(buildInput(req))` | can't be proven; guessing is how a rule earns a FP reputation |

## No options, deliberately

An allowlist would let a project re-approve the dangerous shape wholesale, one config file further from the call site. The remediation *is* naming the fields, and that shape is already silent.

## Scope

mysql2 and better-sqlite3 don't carry the rule — their writes are raw SQL strings, where `no-unsafe-query` already reports. Adding it there would double-report one line from one plugin.

## Test plan

- [x] Factory at **100/100/100/100** (38 tests). Branch closed by exercising a private field (`req.#body`) — the one non-computed property that isn't an `Identifier`.
- [x] `turbo run build test`: 74/74 tasks.
- [x] Barrel locks updated across all five plugins and re-verified.
- [x] All pre-commit and pre-push gates green.
- [x] Regenerated to a fixpoint: `oxlint:shims` → `oxlint:shims:verify --update` (470 → 475), `sync-plugin-stats` (501 → 506), `sync-readmes`, scope audit.

## After merge

Completes W1 (O1–O5). New rules land in `strict` only; promotion waits on W6's measured FP profile per plan §1.6.


ofri-peretz:
All three addressed in `64349b8c` — you were right, and this was a design miss rather than an edge case.

`createMassAssignmentRule` only checked `bindings.size === 0`, which proves a driver import exists *somewhere in the file* and nothing about the receiver. `createUnscopedMutationRule` has carried the `receiverPattern` guard for exactly this since it shipped; I imported `driverBindings` from that same file and did not notice the check sitting next to it.

Your point about `values`/`set` is what makes it bad rather than theoretical — those are Drizzle's *entire* write surface and also `Map.prototype.set`, `Headers.set`, `URLSearchParams.set`, `FormData.set`. Any file importing drizzle that passed `req.body` to a Map reported as database mass assignment.

Added `receiverPattern` to the config plus the receiver-base guard, with per-driver patterns matching the ones `no-unscoped-mutation` already uses — including the exact drizzle pattern you quoted. `this.db.create(...)` still reports, since `receiverBaseName` reads through `this` to the property name; `(a || b).create(...)` does not, because an unnameable receiver cannot be shown to be a handle.

Your suggested `cache.create(req.body)` fixture is in as a `valid` case, alongside locks for the Map, the bound import, the `this.db` chain and the unnameable receiver.

PR #399, opened 2026-08-05
fix(docs): count rules from the rules object, not the whole file

ofri-peretz:
## The number changes

| | before | after |
|---|---|---|
| **Published rules** | 513 | **455** |
| Total incl. unpublished | 541 | 481 |
| Plugins with a wrong count | 21 of 30 | **0** |

⚠️ **This changes every rule count on the docs site, the READMEs and the badges.** Flagging that explicitly — the figure also appears in external material.

## What was wrong

`sync-plugin-stats.ts` derived each plugin's count by running

```ts
content.match(/^\s+'[a-z-]+'\s*:/gm)
```

over the **whole** of `index.ts`. That also matches `plugins: { 'x-security': plugin }` inside every preset, so each plugin was over-counted by roughly one per config it ships. 68 phantom rules across 21 plugins.

Surfaced as a one-line nit on #396 ("MCP says 5, has 2"). It's ecosystem-wide.

## Two errors in the other direction

Scoping the parse to the `rules` object exposed two undercounts the old pattern hid:

- **`eslint-plugin-import-next` writes unquoted keys** — `named: named,`, `default: defaultRule,`. A quoted-only pattern missed all of them: **−7**.
- **`order: enforceImportOrder` is an alias** — a second id for an implementation already counted. The oxlint shim generator has always reported aliases separately (`12 flat + 12 aliased`), so counting distinct implementations is what makes the two sources agree.

## Why nothing caught it

Nothing compared the parse against an independent source. `.agent/oxlint-jsplugins-manifest.json` is one — the shim generator builds it by `require()`-ing each plugin's **built output** and reading `Object.keys(rules)`, so it counts what consumers can actually configure.

**All 30 published plugins now agree with it.** The new lock is exactly that comparison, and it's the only thing standing between a parser tweak and a wrong number on the storefront.

## Test plan

- [x] `plugin-stats-rule-count-lock.test.ts` — 10 tests: the cross-source comparison, plus units pinning the three cases that made this invisible (a config key masquerading as a rule, unquoted keys, an alias).
- [x] Guards against a vacuous pass — asserts both catalogs parse non-empty before comparing.
- [x] Full `docs` suite: 20/20 tasks. Nothing else asserted the old numbers.
- [x] `sync-readmes` re-run so the badges match.

### DIFF
--- a/.changeset/fix-plugin-stats-rule-count.md
+++ b/.changeset/fix-plugin-stats-rule-count.md
@@ -1,0 +1,29 @@
+---
+'docs': patch
+---
+
+Correct the published rule count. It was overstated by 58 rules (513 → 455).
+
+`sync-plugin-stats.ts` derived every plugin's rule count by regex-matching
+`/^\s+'[a-z-]+'\s*:/gm` against the whole of `index.ts`. That also matched
+`plugins: { 'x-security': plugin }` inside every preset, so each plugin was
+over-counted by roughly one per config it ships — 21 of 30 published plugins,
+68 phantom rules.
+
+Fixing the scope surfaced two errors in the other direction:
+
+- `eslint-plugin-import-next` writes its keys unquoted (`named: named,`), which
+  a quoted-only pattern missed entirely — undercounting it by 7.
+- `order: enforceImportOrder` is an alias, a second id for an implementation
+  already counted. The oxlint shim generator has always treated aliases that way
+  ("12 flat + 12 aliased"); counting distinct implementations makes the two
+  agree.
+
+These numbers feed `interlace-numbers.json`, which is the single source for
+every count on the docs site, the READMEs and the badges.
+
+Nothing caught this because nothing compared the parse against an independent
+source. `.agent/oxlint-jsplugins-manifest.json` is one — the shim generator
+builds it by `require()`-ing each plugin's built output and reading
+`Object.keys(rules)`. All 30 published plugins now agree with it, and a new lock
+asserts exactly that comparison.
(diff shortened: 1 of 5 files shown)


ofri-peretz:
All three fixed in `f58a8dbf`. The second one was a real defect in my fix, not a hypothetical.

**Word boundary** — `/export const rules[^=]*=\s*\{/` matches `export const rulesHelper = {`, so a file declaring one first would have its helper counted and the real map never read. That's a confident wrong number rather than a visible failure, which is the worst kind here. Anchored with `\b` plus an explicit optional type annotation.

**Factory-call dedup** — you're right, and it was live-fire rather than theoretical:

```ts
'rule-a': makeRule({ foo: 1 }),
'rule-b': makeRule({ bar: 2 }),
```

both resolved to `makeRule` and counted 1. Only a **bare identifier** value can be an alias; a call is always its own rule. Call-valued entries are now counted individually and never deduplicated.

Narrowing that immediately exposed a third defect the first version had been hiding: the one real alias in the ecosystem is written

```ts
order: enforceImportOrder, // Alias for backwards compat with eslint-plugin-import
```

and the trailing comment made the value read as an expression rather than a bare identifier — silently turning the alias back into a second rule. Caught it because the cross-source check went from ✅ to `import-next 56 vs 55`. Comments are stripped before classifying now.

**Tests** — all four cases you'd want: two rules from one factory, an alias with a trailing comment, a same-named earlier export, and a type-annotated `rules` export. 14 tests total.

Count is 456 rather than 455 because main gained `no-tool-description-injection` when #396 merged. All 30 published plugins still agree with the oxlint runtime manifest.

PR #414, opened 2026-08-06
chore(workspace): delete the renamed pg and jwt plugin sources

ofri-peretz:
## Why this is not just cleanup

Both packages were renamed to their `-security` names and every published version is deprecated on npm. But the sources stayed in `packages/`, and `.changeset/config.json` has `ignore: []` — so **every release versioned and republished them**:

```
eslint-plugin-pg@1.4.13    published 2026-08-05
eslint-plugin-jwt@2.2.13   published 2026-08-05
```

A freshly published version carries no deprecation flag. So each release silently **un-deprecated** both packages, and `npm i eslint-plugin-pg` resolved to a clean `latest` with no warning until someone re-ran `npm deprecate`.

Deleting the sources is what stops the loop. Re-deprecating alone gets undone by the next release.

## Nothing published is lost

- The `-security` packages carry **identical rule sets** — 13 each, compared directory by directory.
- They keep the original `pg/` and `jwt/` **rule namespaces**, so no consumer config changes.
- Published catalogue is **unchanged: 465 rules across 30 plugins**. Only the including-unpublished count drops (491 → 465), because the removed entries were already flagged unpublished.

## A user-facing bug this surfaced

The playground's copy-config button derived package names as `eslint-plugin-<prefix>`. Since the rename deliberately kept the namespaces, `jwt/` and `pg/` findings emitted install lines for the **deprecated** packages on the public site. Both prefixes are now mapped explicitly, with a lock.

## Test plan

- [x] `turbo run test typecheck` — 86/86.
- [x] `test:scripts` — 1026 passed. Two locks caught real fallout rather than needing to be silenced: `oxlint-export-lock` found `pg` / `jwt` still named in benchmark `--plugins` lists, where they were being **silently dropped at runtime**; `playground-lock` found the deprecated install lines.
- [x] `validate-docs` 10/10, `lint:md` 0 issues across 828 files, oxlint clean.
- [x] Catalogs regenerated: oxlint shims, scope audit, plugin-stats.

## Blast radius

Registries in `validate-docs`, `sync-rule-docs`, `ilb-plugin-scope-audit`, `ecosystem-integrity` and the README tooling lose their two entries. Benchmark plugin shorts move to `postgresql-security` / `jwt-security`. Synthetic fixtures in the ci-shard and lazy-barrel locks repoint to surviving packages.

## After merge

The deprecated packages stay on npm exactly as they are — deprecation is a registry flag, unaffected by source removal — and can no longer be accidentally republished. Verify with `npm view eslint-plugin-pg deprecated` after the next release: it should still return the notice.


ofri-peretz:
## The loop ran again while this PR was open

Resolving this branch against `main` conflicted exactly where the PR predicts — the release bumped versions *inside the packages being deleted*. Checking npm after that release:

```
eslint-plugin-pg    26 versions, latest=1.4.14, NOT deprecated: [1.4.14]
eslint-plugin-jwt   23 versions, latest=2.2.14, NOT deprecated: [2.2.14]
```

Both were fully deprecated yesterday. The release published `1.4.14` and `2.2.14`, both **undeprecated**, and both are now `latest` — so `npm i eslint-plugin-pg` resolves to a version carrying no notice, for the second time this week.

That is the argument for this PR, demonstrated rather than asserted: re-running `npm deprecate` does not hold while the sources sit in `packages/`.

**After merging, the two new versions still need one `npm deprecate` pass** to clean up what this release published. After that it stays clean, because there is nothing left to republish.

Also fixed here: `secure-coding`'s CVSS-docs consistency lock imported rules from both packages by relative path, so the file failed to load once they were gone. The rename kept the `jwt/` and `pg/` namespaces, so the imports move to the `-security` packages and every assertion holds unchanged.

Catalogs regenerated — published totals unchanged at **465 rules across 30 plugins**; the including-unpublished count now matches, since the duplicate entries are gone.

ofri-peretz:
Parking this briefly to free the CI runner pool — the account is at its concurrency cap and PRs are timing out while queued rather than failing. Working #409 to green first, then reopening these one at a time in order: #414, #427, #422, #417, #416, #426.

No code change; the branch is untouched.

PR #426, opened 2026-08-06
chore(eslint-devkit): ratchet against regexing printed source

ofri-peretz:
Follow-up to #422, which fixed one instance of this. The audit found **205 more, across 61 files**.

## The defect class

A rule that runs `.test()` / `.includes()` / `.match()` over `sourceCode.getText(node)` is guessing. Printed text carries identifiers, comments and whitespace — everything the parser already separated out — so a match says nothing about structure.

Both failure directions have shipped in this repo:

| direction | instance |
|---|---|
| **False positives** | `no-xpath-injection` matched `render.text() + input` (an identifier) and a `/* //user[@id] */` **comment** as XPath construction — fixed in #422 |
| **Self-suppression** | `express-security/require-route-authentication` matched `require(...)` against its auth-middleware list, so `app.use(require('body-parser'))` marked the file globally authenticated and **silenced every route in it** (#313) |

The second is the dangerous one: the rule scores a perfect false-positive rate while protecting nothing.

## What this adds

`npm run audit:gettext` counts the sites and **fails when the number grows**. The 205 existing sites are **baselined, not fixed** — this stops the class expanding while it gets burned down, the same shape as the existing portability baseline. Wired into the aggregate quality gate.

**No rule behaviour changes.**

## Test plan

- [x] Ratchet verified in both states: passes at baseline, and **fails on an injected violation** (`packages/eslint-plugin-node-security/src/rules/no-ssrf/index.ts: 0 → 1`).
- [x] 7 unit tests on the detector, covering both shapes that have caused real defects — the inline call result, and a variable assigned from `getText` and classified later.
- [x] Negative cases locked too: `getText` used for **reporting**, for **fixers**, and classification of an AST value like `node.value` are all deliberately not flagged.
- [x] `test:scripts` 1083 passed, `turbo run typecheck` 53/53, markdown and oxlint clean.

## Burning it down

The baseline is a work-list, not a target. Each site should become AST analysis — literal values, node types, member paths. The highest-priority ones are those where a match *suppresses* a report rather than causing one, since those fail silently; `no-clickjacking` (`testText.includes('top != self')`) and the two `no-missing-cors-check` copies are that shape.

## Summary by CodeRabbit

- **New Features**
  - Added automated auditing to detect increases in gettext classification patterns.
  - Added a baseline containing the current 205 tracked findings across 61 files.
  - Added a command to generate or update the audit baseline.

- **Quality Improvements**
  - Integrated the gettext audit into the standard quality-check workflow.
  - Added automated coverage for supported classification patterns, exclusions, and baseline consistency.
  - The audit reports reductions and fails when new findings are introduced.


ofri-peretz:
Parking this briefly to free the CI runner pool — the account is at its concurrency cap and PRs are timing out while queued rather than failing. Working #409 to green first, then reopening these one at a time in order: #414, #427, #422, #417, #416, #426.

No code change; the branch is untouched.

PR #474, opened 2026-08-10
fix(ci): declare the docs build's Supabase env so turbo's cache key includes it

ofri-peretz:
## Summary

Two config defects, found when #472's production deploy failed. Together they meant the docs site could not ship the canonical download figures.

**1. `.vercelignore` patterns were unanchored.**

```
scripts        →  matches apps/docs/scripts too
tools          →  matches apps/*/tools too
packages, playground, distribution — same
```

`apps/docs/scripts/sync-plugin-stats.ts` is the first thing `docs#build` runs, and it was being stripped from every upload. It is committed and not gitignored, so nothing pointed at the cause.

This has been latent for as long as the patterns have existed. Deploys kept passing because `docs#build` kept hitting the turbo cache and never executed the command — the first deploy that genuinely rebuilt docs died with `ERR_MODULE_NOT_FOUND`. A leading slash restores the original intent: exclude the repo-root directories, not every namesake beneath them.

**2. The docs build env was never declared to turbo.**

The build log has been saying so:

> the following environment variables are set on your Vercel project, but missing from `turbo.json`. These variables **WILL NOT** be available to your application

`SUPABASE_URL` and `SUPABASE_ANON_KEY` are exactly that. `/stats` would have built with no data source and silently omitted the download card — #472's graceful degradation working correctly while hiding a config bug, which is the worst possible combination.

## Test plan

- [x] `turbo run test --filter=...[origin/main]` passes (the `"// env"` comment key I first tried is rejected by turbo's schema — removed).
- [x] `prettier --check turbo.json` clean.
- [x] Full pre-push gate green.
- [ ] Post-merge: production deploy reaches Ready and `/stats` shows **360,361** with the "since December 2025" label.

## Summary by CodeRabbit

* **Chores**
  * Updated build configuration to recognize the required Supabase environment variables during builds.


ofri-peretz:
Resolved the conflict against `main`.

The `.vercelignore` half of this PR is **obsolete** — #476 landed a broader fix that deletes those files entirely rather than anchoring the patterns, so I took `main`'s deletion.

What remains is the `turbo.json` half, which is still needed: `SUPABASE_URL` / `SUPABASE_ANON_KEY` are read at build time by `apps/docs/.interlace/lib/scorecard-data.ts`, and neither is declared in `build.env` on `main`. Undeclared, they are absent from the cache key, so rotating a key can serve a stale cached build — the same failure mode as the `content/` inputs gap.

Retitling would be fair since the headline no longer matches the content.

PR #488, opened 2026-08-10
fix(bench): clone the repo the headline suite actually benchmarks

ofri-peretz:
## What broke

The first weekly benchmark after #454 failed before rendering a single badge:

```
Repo not found: /home/runner/work/eslint/eslint/oos/nestjs
##[error]Process completed with exit code 2
```

Run: [31378731853](https://github.com/ofri-peretz/eslint/actions/runs/31378731853), 2026-08-10 10:20 UTC.

## Why

Two lists have to agree, and nothing links them:

| List | Contents | Set by |
|---|---|---|
| corpus clones | next.js, supabase, lodash, shadcn-ui, payload, vercel-ai | pinned to the flagship manifest |
| repos benchmarked | `${{ inputs.headline_repo \|\| 'nestjs' }}`, shadcn-ui | ILB-Headline steps |

`nestjs` is the headline suite's Node corpus. No manifest references it, so the clone step — explicitly "pinned to the same repos the manifest references" — never cloned it.

This was invisible while `three.js` was in the list: the headline default resolved to three.js, which *was* cloned. #454 removed three.js as a corpus we must not publish, which was right, and left the headline suite with no repo at all.

**Consequence:** the published badges still read `corpus: three.js`. Not because the fix didn't land, but because the run that would replace them exits 2 before the badge step. Every badge on the README has been serving the pre-#454 measurement since.

## The fix

- `clone nestjs https://github.com/nestjs/nest.git`.
- A preflight asserting every repo about to be benchmarked is in the corpus. The old failure was loud but late — it surfaced after cloning six large repos and building 31 packages. The preflight also catches a `workflow_dispatch` whose `headline_repo` input nobody clones.

## Verification

| Case | Expected | Result |
|---|---|---|
| nestjs absent (what happened) | blocks, names it | ✅ `missing: nestjs`, exit 1 |
| nestjs present (after fix) | passes | ✅ |
| dispatch names an uncloned repo | blocks, names it | ✅ `missing: vue-core`, exit 1 |
| `lint:workflows` | green | ✅ 34 files |

The OOS cache restores via `restore-keys`, so an old cache without nestjs self-heals: the clone helper clones any missing directory.

## After merge

Badges stay stale until a run succeeds — the next cron, or a manual dispatch. Dispatch publishes to production Pages and pushes to main, so that call is the maintainer's.

## Summary by CodeRabbit

* **New Features**
  * Weekly benchmark runs now include the NestJS repository alongside the existing benchmark repositories.
  * Added preflight validation to confirm required benchmark repositories are available before execution.

* **Bug Fixes**
  * Benchmark runs now stop early with a clear error when required repositories are missing.

### DIFF
--- a/.github/workflows/weekly-benchmark.yml
+++ b/.github/workflows/weekly-benchmark.yml
@@ -75,6 +75,26 @@ jobs:
     # nothing here is rushed into a misleading fast number.
     timeout-minutes: 90
 
+    env:
+      # THE corpus. One list, read by the clone step and by the preflight that
+      # validates what we are about to benchmark — because the defect this
+      # replaced was two lists that had to agree with nothing linking them:
+      # the clones were pinned to the flagship manifest, ILB-Headline
+      # benchmarks nestjs, no manifest cites nestjs, so it was never cloned.
+      # Anything benchmarked must appear here.
+      CORPUS: |
+        next.js   https://github.com/vercel/next.js.git
+        supabase  https://github.com/supabase/supabase.git
+        lodash    https://github.com/lodash/lodash.git
+        shadcn-ui https://github.com/shadcn-ui/ui.git
+        payload   https://github.com/payloadcms/payload.git
+        vercel-ai https://github.com/vercel/ai.git
+        nestjs    https://github.com/nestjs/nest.git
+      # Both headline runs, as one list to validate against CORPUS. The second
+      # is hardcoded to shadcn-ui in its step; keep them in step.
+      HEADLINE_REPO: ${{ github.event.inputs.headline_repo || 'nestjs' }}
+      FRONTEND_REPO: shadcn-ui
+
     steps:
       # Pages is a single production surface and the commit step writes to
       # main, so a dispatch from a feature branch would publish that branch's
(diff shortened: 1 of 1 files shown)


ofri-peretz:
All three review findings were correct and are fixed in `06ea6f9c` — as a replacement for the first patch, not an addition to it.

Finding 2 was the one worth stopping on: the preflight ran *after* the clone step, so it did not do the thing this PR's description claimed it did. That reframed the other two. Adding a seventh `clone` line fixed the symptom and left the shape — two lists that must agree, with nothing linking them — exactly as it was.

`CORPUS` is now a single job-level list that the clone loop and both guards read.

| Finding | Fix |
|---|---|
| Preflight after cloning | Moved before the cache and clone steps; validates against `CORPUS`, costs seconds |
| Cache key omits corpus changes | `oos-corpus-v2-`, hashing this workflow file, since `CORPUS` lives in it |
| `[ -d ]` accepts a non-checkout | `git rev-parse HEAD` per benchmarked repo, printing each SHA |

Each guard was run against the state it exists to catch:

```
preflight   nestjs declared            -> pass
            uncloned dispatch input    -> BLOCK: vue-core
            CORPUS without nestjs      -> BLOCK: nestjs      (the original bug)
HEAD check  bare dir, no .git          -> BLOCK              (what `test -d` passed)
            real checkouts             -> pass, prints SHAs
            .git present, HEAD broken  -> BLOCK that repo only
clone loop  parses all 7 name/url pairs off CORPUS
```

`lint:workflows` green across 34 files.

On the cache: `restore-keys` still gives a warm partial restore on the v2 key change, and the clone helper refreshes what it finds — so the one-time re-save is the whole cost.

PR #530, opened 2026-08-12
fix(modularity): no-external-api-calls-in-utils requires a resolved HTTP client

ofri-peretz:
## Summary

- `no-external-api-calls-in-utils` matched on the **method name** alone, so every `Map` / `Set` / `Headers` / `URLSearchParams` / `Cache` lookup in a `utils`/`lib`/`helpers` file was reported as a network call. The 2026-08-11 dogfooding sweep (123 files) produced **45 findings across 17 files**, and every sampled one was a cache lookup — `store.get(key)`, `asOfById.get(def.id)`, `series.get(def.id)`.
- A call is now reported only when its callee **resolves to an HTTP client**:
  - `fetch` / `window.fetch` (or `globalThis`/`global`/`self`),
  - a binding imported or required from a module in the new `httpModules` option (axios, got, ky, node-fetch, undici, superagent, request, phin, needle, `http(s)`, `node:http(s)`, `http2`),
  - an alias of one — `const api = axios.create()` — resolved to a fixpoint on `Program:exit`, so the client may be declared *after* its use,
  - an explicit `object.method` pair in `networkMethods` — the escape hatch for in-house clients the import tracking can't see.
- Bare module names still seed the client set, so `axios.get(...)` is caught without a visible import — **except** `request` and `got`, which collide with Express's `request` object (`request.get('host')`) and ordinary English.
- Reporting moved to `Program:exit` (calls are collected, then decided once bindings are known); the synthetic-branch coverage test now drives that listener.
- `KNOWN-LIMITATIONS.md` documents the new contract, the two new FN classes (multi-hop aliases, unresolved receivers) and the one new FP class (a local variable named after an HTTP module).

## Test plan

- [x] `npx vitest run` in `packages/eslint-plugin-modularity` — **149 passed**, coverage **100 / 100 / 100 / 100** (the package gate).
- [x] Regression lock in [no-external-api-calls-in-utils.test.ts](packages/eslint-plugin-modularity/src/tests/no-external-api-calls-in-utils.test.ts) — a `built-in collections are not network calls` block asserts silence on `new Map().get/set/delete`, `new Headers().get`, `new URLSearchParams().get/delete`, `new Set().add`, `sessionStorage.getItem`, `request.get('host')` and `getStore().get(key)`; it fails on the pre-fix rule (every one of those was an error) and passes on the fix.
- [x] Genuine positives kept, as invalid cases: imported `axios`/`undici`/`got` (ESM + `require`), aliased `axios.create()` used before declaration, `window.fetch`, and the `networkMethods: ['httpClient.get']` escape hatch.
- [x] `npx tsc -p packages/eslint-plugin-modularity/tsconfig.lib.json --noEmit` — clean.
- [x] `npm run ilb:scope-audit`, `npm run check-stale-artifacts`, `npm run oxlint:shims:check`, `markdownlint-cli2` — clean.
- [x] Corpus measurement: linted 49 real `utils`/`lib`/`helpers` files across `apps/docs`, `~/repos/agents` and the serverless monorepo with the old rule and the new one — **47 findings → 13**. Every survivor is a real `fetch(...)`; every dropped finding is a collection lookup (`descriptions.get(...)`, `memoryCache.delete(...)`, `u.searchParams.get('utm_source')`).

Pre-push hooks `typecheck`, `portability` and `build-and-test` were excluded on the push: all three fail on this worktree for environment reasons unrelated to this diff — a stale `@interlace/eslint-devkit` build in the shared `node_modules` (`createPayloadResolver` etc. missing, in browser-security / sequelize / mysql / prisma), `oxlint not installed at node_modules/oxlint/dist — run npm ci`, and `docs#build` failing on Next.js workspace-root inference. CI is the gate for those.

## After merge

`eslint-plugin-modularity` gets a patch bump via the changeset. Users of the `modularity` preset should see the rule go quiet on cache-heavy utility modules while still flagging `fetch`/axios calls. This is the same root cause as the `express-security/no-missing-security-headers` FP on `perDay.set(` — that one ships separately; the rest of the ecosystem is worth grepping for detectors keyed on a bare `.get(` / `.set(` / `.post(` / `.delete(`.


ofri-peretz:
All three addressed in `1f821e43`.

**1. The `KNOWN-LIMITATIONS.md` claim was wrong — you were right.** I checked the
`Program:exit` fixpoint rather than taking it on the doc's word, and then wrote
the test to be sure:

```ts
const api = axios.create();
const api2 = api.create();
api2.get(url);          // reports
```

The `while (changed)` loop propagates client-ness until nothing new is added, so
alias *depth* was never the limit. Rewrote the entry to name the actual failure —
opacity: `const c = getClient(); c.get(url)`, where `rootName()` resolves the
callee to `getClient`, which is not a known client, so `c` never joins the set.
The doc now states the fixpoint handles depth, so this cannot quietly regress
into folklore again.

That transitive case is now an `invalid` fixture. A doc claiming a false negative
that isn't real is its own kind of defect — it invites someone to "fix" working
code.

**2. Namespace-import test added**, using your case verbatim. It passes, which
confirms the path worked and was simply unpinned.

**3. `networkMethods` description added.** Slightly expanded to make the
distinction explicit, since the surprising part is that a dotted entry bypasses
client resolution altogether:

> HTTP method names to flag. A bare name (`"get"`) applies only when the receiver
> resolves to a known HTTP client. A dotted entry (`"httpClient.get"`) is an
> exact `object.method` escape hatch that bypasses client resolution entirely —
> use it for an in-house client this rule cannot trace.

35 tests, 100% statement/branch/function/line coverage across the package.

PR #531, opened 2026-08-12
fix: open-issue sweep — plugin prefix identity, logging FNs, CI gates

### LINKED ISSUE #504: fix(node-security): no-timing-unsafe-compare fires on string-literal and boolean comparisons
**106 findings** in the 8-repo scan — the largest remaining false-positive source.

## Repro

```js
if (revokedToken === 'access') { }   // ← "timing attack"
if (prevAuth === curAuth) { }        // ← two booleans
```

Both report `CWE-208 | Using === to compare secrets enables timing attacks`.

`'access'` is a **string literal sitting in the source file** — there is no secret on either side. The rule's evidence is the identifier name containing `token` / `auth`. Real site: `okta/okta-auth-js` `lib/oidc/dpop.ts:185`, and `AuthStateManager.ts:44` where it flags `JSON.stringify(a) === JSON.stringify(b)` used for change detection.

## Fix direction

A timing-unsafe comparison needs **both operands to be plausibly secret**. A comparison against a literal, a boolean, or a serialized-state string cannot leak a secret that isn't already in the source. Require the other operand to be non-literal, and drop name-only evidence.

## Acceptance

- TDD: the three shapes above added as `valid` and failing before the fix
- Existing true positives (comparing two secret-bearing identifiers) still report
- Re-run `agents/footprint/scripts/scan-targets.ts` — this rule's count should fall to near zero on the 8 targets

### LINKED ISSUE #505: fix(secure-coding): no-redos-vulnerable-regex flags anchored linear patterns
**61 findings.** Identified 2026-08-09 as fix-order item #8 — now the oldest known unfixed defect.

## Repro

```js
const V3_URL_REGEX = /^https:\/\/js\.stripe\.com\/v3\/?(\?.*)?$/;
```

Reports `Nested Quantifier Pattern: exponential backtracking | CRITICAL`.

Anchored at both ends, one optional group, **no nested quantifier**. This is linear.

Real site: `stripe/stripe-js` `src/shared.ts:23-24`. Stripe was the cleanest repo in the whole scan — 2 findings in 79 files — and both were this.

## Fix direction

A nested quantifier is a quantified group that itself contains a quantifier: `(a+)+`, `(a*)*`, `(a|aa)+`. `\/?` followed by a separate `(\?.*)?` is two independent optional groups, not nesting. The check needs to look at quantifier *nesting depth* in the parsed pattern, not at the count of quantifier characters.

## Acceptance

- TDD: the Stripe regex as `valid`, failing first
- Genuine catastrophic patterns (`(a+)+$`, `(\w+\s?)*$`) still report
- Re-run the corpus scan

### LINKED ISSUE #506: fix(secure-coding): no-sensitive-data-exposure matches the word 'password' in prose
**41 findings.**

## Repro

```js
console.warn(
  "[WARN] Redis server does not require a password, but a password was supplied."
);
```

Reports `CWE-532 | Sensitive data detected in logs: password | HIGH`.

The word *password* appears in an English sentence. **No value is logged.** Real site: `redis/ioredis` `lib/redis/event_handler.ts:271` and `:279`.

## Fix direction

A static string literal with no interpolation cannot leak a runtime secret — whatever words it contains. Reporting should require an interpolated expression or a member access that carries the value (`user.password`, `${token}`), not a word in a constant.

Note `literalCarriesSecret()` already exists in this rule for a related case (requiring `\b<word>\b[^:=\n]{0,24}[:=]\s*\S`); the logging path does not use it.

## Acceptance

- TDD: the ioredis warning as `valid`, failing first
- `console.log(user.password)` and `` console.log(`token=${t}`) `` still report

### LINKED ISSUE #507: fix(jwt-security): no-decode-without-verify matches any method named decode
**31 findings.**

## Repro

```js
file.content = file.decode(raw)   // Shopify/cli — a TOML file parser
const accessJwt = sdk.token.decode(accessToken)   // okta — Okta's own method
```

Both report `JWT decoded without verification`. Neither involves `jsonwebtoken`.

The one case that *is* a real `jwt.decode` is `twilio/twilio-node` `TokenAuthStrategy.ts:49` — and Twilio documented why on the line above:

```js
// Decode the token without verifying the signature, as we only
// want to read the expiration for this check
const decoded = jwt.decode(token) as JwtPayload;
```

Reading `exp` to decide when to refresh your own token is standard; the signature is verified server-side.

## Fix direction

Require the callee to resolve to an imported JWT library (`jsonwebtoken`, `jose`, `jwt-decode`) — the SDK-evidence gate this repo already applies elsewhere. A bare method named `decode` is not evidence of anything.

Separately, consider whether the Twilio shape (decode → read `exp` only, no trust decision) warrants a lower severity or an exemption.

## Acceptance

- TDD: the TOML and `sdk.token.decode` cases as `valid`, failing first
- `import jwt from 'jsonwebtoken'; jwt.decode(t)` with no `verify` still reports

### LINKED ISSUE #512: feat(node-security): non-crypto-use exemption for hash rules
Not a false positive — a rule-design gap.

## Repro

```js
// redis/ioredis lib/Script.ts:15
this.sha = createHash("sha1").update(lua).digest("hex");
```

Reports `CWE-327 | Use of weak hash algorithm: SHA-1 | CRITICAL`.

SHA-1 **is** used, so the detection is correct. But this is the **EVALSHA script identifier the Redis wire protocol mandates** — a cache key, not a security control. A maintainer reading this finding correctly concludes the tool does not understand their code.

The same shape will hit ETags, content-addressed caches, and cache-busting hashes.

## Fix direction

An exemption path for non-security hashing. Options, roughly in order of preference:

1. A recognised set of non-crypto uses (assignment to `sha`/`etag`/`cacheKey`, `digest('hex')` fed to a Map key)
2. An `allowNonCryptographic` option, defaulting off, documented with the Redis case
3. A JSDoc annotation (`@non-crypto-hash`), consistent with the `@safe` annotations elsewhere

Whichever is chosen, the message should name the distinction so a reader can tell which case they are in.

### LINKED ISSUE #514: fix(jwt-security): published package crashes ESLint — recommended emits jwt/ prefixes after rename
**Severity: the plugin does not run at all.** Worse for an adopter than any false positive.

## Repro

```js
import jwtSecurity from 'eslint-plugin-jwt-security';
export default [{
  plugins: { 'jwt-security': jwtSecurity },
  rules: { ...jwtSecurity.configs.recommended.rules },
}];
```

```
A configuration object specifies rule "jwt/no-algorithm-none",
but could not find plugin "jwt".
```

Verified against the package published on npm. Hit while building `agents/footprint/scripts/scan-targets.ts`, which now derives plugin keys from the rules themselves to work around it.

## Root cause

The source package is still `packages/eslint-plugin-jwt`. It **publishes** as `eslint-plugin-jwt-security`, but `configs.recommended` still emits the pre-rename `jwt/` prefix. Registering under the package name — the obvious thing — fails; only registering as `jwt` works.

`postgresql-security` is suspected to have the same shape (`pg/` prefix); not yet verified, the package would not resolve locally.

## Why this is not a one-line change

Switching to `jwt-security/` **breaks every adopter currently registering as `jwt`**, which is the only registration that works today. It needs a release decision first:

- dual prefixes for a deprecation window, or
- a major with a migration note

Plus the source directory rename, the docs slug (see the docs↔package coupling constraint), and a regression lock asserting each plugin's `configs.recommended` rule prefixes match its published package name.

## Acceptance

- Registering by package name works for every published plugin
- A test iterates all plugins and asserts prefix == package suffix

### LINKED ISSUE #515: fix(secure-coding): no-directive-injection flags the correct DOMPurify call
The rule reports the recommended safe pattern:

```js
DOMPurify.sanitize(html, { ALLOWED_TAGS, ALLOWED_ATTR })
```

This is the configured, allow-listed sanitizer call — the answer the rule should be steering people toward. Reporting it tells a developer who did the right thing that they did the wrong thing.

Carried over from the FP worklist in `DISTRIBUTION-PLAN.md`; not yet reproduced in isolation.

## Acceptance

- TDD: the call above as `valid`, failing first
- A genuinely unsafe directive (`sanitize(html, { ADD_TAGS: ['script'] })`, or an unsanitized assignment) still reports

### LINKED ISSUE #516: fix(secure-coding): no-hardcoded-credentials allowInTests is not applying
17 of 18 findings were fixtures in `integration/auth.test.js` — a file `allowInTests` exists to skip.

Either the option is not being read, or its pattern does not match `integration/*.test.js`.

## Acceptance

- Determine which of the two it is before changing anything
- TDD: a credential-shaped literal in `integration/auth.test.js` as `valid` with the default options
- Real credentials in production paths still report

ofri-peretz:
## Summary

One PR for the open-issue backlog. Every number below is **measured** against the
8-repo corpus scan (`okta/okta-auth-js`, `auth0/express-openid-connect`,
`stripe/stripe-js`, `twilio/twilio-node`, `redis/ioredis`,
`paypal/paypal-checkout-components`, `okta/okta-signin-widget`, `Shopify/cli`)
with the local build — before and after — not quoted from the issue text.

Closes #504
Closes #505
Closes #506
Closes #507
Closes #512
Closes #514
Closes #515
Closes #516

## Measured effect on the corpus

| Rule | Before | After | |
|---|--:|--:|---|
| `node-security/no-timing-unsafe-compare` | 106 | **27** | −79 |
| `secure-coding/no-redos-vulnerable-regex` | 61 | **35** | −26 |
| `node-security/no-weak-hash-algorithm` | 7 | **6** | −1 |
| `jwt-security/no-decode-without-verify` | 1 | **2** | +1 — a **new true positive** the jose fix uncovered |
| **Total** | **993** | **888** | **−105** |

Survivors were inspected rather than assumed. What remains for the timing rule
is `password !== confirmPassword`, `hash !== token.claims.at_hash`,
`claims.nonce !== nonce`; for ReDoS it is NFA-confirmed link-header and
URL-validation regexes.

`no-hardcoded-credentials` shows no movement here **because this scan already
excludes test paths** — the 17-of-18 figure in #516 came from a scan that did
not. Its fix is verified by fixtures instead, not by this table.

## The fixes

**#514 — plugins were not usable under their own package name.** A shipped
crash. `eslint-plugin-jwt-security` and `eslint-plugin-postgresql-security` were
renamed but their presets kept emitting the pre-rename `jwt/` and `pg/`
prefixes, so the shape every README shows failed outright. Legacy keys stay
registered for a deprecation window. Exactly 2 of 30 plugins were affected; the
`pg` case was "suspected, not verified" in the issue and is now verified.

**#504 — comparisons against constants.** `revokedToken === 'access'`
(`okta/okta-auth-js` `lib/oidc/dpop.ts:185`) cannot leak a secret: the value
compared against is in the source. Three guards — constant operands, boolean
predicate names (`isAuthenticated`), and namespaced constants
(`AUTHENTICATOR_KEY.WEBAUTHN`, 73 of the 88 left after the first two). The last
guard requires **both** a namespace-cased object and a constant-cased property
on a non-computed member, so `credentials.API_TOKEN`, `secrets[API_TOKEN]` and
`process.env.API_TOKEN` (both spellings) all still report.

**#505 — the NFA verdict is now final, and invalid syntax is not ReDoS.**
`checkWithScslre` returned a boolean, so "analysed and safe" and "could not
analyse" were the same value; every pattern the NFA cleared fell through to
regexes-matching-regex-source and got reported anyway. That is how Stripe's
`/^https:\/\/js\.stripe\.com\/v3\/?(\?.*)?$/` became `CRITICAL`. Stripe: 3 → 0.
Separating parse failure from analyser failure then left the heuristic layer
with nothing to do but overrule clean verdicts, so it is deleted.

**#516 — the diagnosis was neither option in the issue.** The option is read
and its patterns *do* match `integration/*.test.js`. It defaulted to `false`,
and `configs.recommended` registers the rule as bare `'error'` with no options —
so the exemption was never on. Default flipped; production paths unchanged.

**#512 — non-cryptographic hashing.** `this.sha = createHash('sha1')…` in
`redis/ioredis` is the EVALSHA identifier the Redis protocol mandates. Narrow by
construction: it is about where the value *lands*, and seven invalid fixtures
pin that "call it `sha`" cannot silence the rule on a real security control.

**#507 and #515 — the reported FPs no longer reproduce.** Both are locked as
`valid` fixtures taken from the real files. Verifying them surfaced three
genuine defects, all fixed: jose's `decodeJwt` went undetected (the method set
listed `decodeJWT`, a spelling no library ships — this is the +1 true positive
above); the foreign-import gate never checked chained receivers like
`sdk.token.decode`; and `DOMPurify.sanitize(html, { ADD_TAGS: ['script'] })` was
**silently accepted** — markup that executes while reading as sanitized at the
call site.

## Also fixed

**The pre-push gate could not go green in a fresh worktree.**
`oxlint:shims:verify` requires `dist/` for all 30 plugins, but the build ran
under `--filter=...[origin/main]`. Any push not touching every plugin failed on
`Cannot find module`. Build is now unfiltered; `test` keeps the filter.

## Code scanning

3 of 26 alerts are fixable in code, and all 3 are fixed: two unpinned actions
pinned to SHA, and `weekly-benchmark.yml` top-level `permissions:` reduced to
`contents: read`. A stale `# v5` comment on a v6 SHA was corrected.

The other 23 cannot be closed by a PR: 12 are job-scoped `write` permissions the
jobs require, 6 are already version-pinned installs where Scorecard wants
hash-pinning, and 5 are repo-level signals (Fuzzing, SAST, Security-Policy,
Code-Review, CII) needing OSS-Fuzz, CodeQL enablement and a review policy.

## Test plan

- [x] Every fix mutation-verified — reverting each rule turns its new fixtures red
- [x] `turbo run build test` across the whole graph: **70/70 tasks green**
- [x] Every touched rule at 100% statement/branch/function/line coverage
- [x] `benchmarks/__tests__/plugin-prefix-identity.lock.test.ts` — 62 tests across all 30 plugins, verified red on the unfixed code, with a guard-the-guard on the plugin count
- [x] `scripts/__tests__` 1106 tests green; `lint-workflows.ts` clean across 34 workflows
- [x] `ilb:scope-audit` 2 findings → 0
- [x] Rule docs updated and generated MDX re-synced

## Review

Three of four CodeRabbit findings were real defects and are fixed — see the
[response comment](https://github.com/ofri-peretz/eslint/pull/531#issuecomment-5263645446).
The fourth (splitting benchmark execution from publication) is a pre-existing
workflow redesign, declined here with a reason.

## Repo setting to change

Settings → Actions → General → **Fork pull request workflows from outside
collaborators**. Until this changes, release PRs stall in `action_required`.


ofri-peretz:
## Review responses

Three of the four findings were real defects and are fixed in `0a2a2b4`. The fourth is declined with a reason.

### ✅ `isNamedConstant` infers a constant from an arbitrary member name — **fixed**

Correct on all three counts, and the third one stings: `secrets[API_TOKEN]` is the same computed-property bug this PR already fixes twice elsewhere (`no-sensitive-data-exposure`, `no-weak-hash-algorithm`), reintroduced in new code.

Rather than special-case each hole, the guard now requires **both** halves to carry the convention — a namespace-cased object (PascalCase or SCREAMING_SNAKE) **and** a constant-cased property, on a non-computed member:

```js
if (node.computed) return false;
if (node.object.type !== AST_NODE_TYPES.Identifier) return false;
if (node.property.type !== AST_NODE_TYPES.Identifier) return false;
return NAMESPACE_NAME.test(node.object.name) && NAMED_CONSTANT.test(node.property.name);
```

All 73 corpus findings the guard exists for satisfy both halves, so nothing is lost. The three named cases now report:

| Shape | Why it reports |
|---|---|
| `userToken === credentials.API_TOKEN` | `credentials` is camelCase — a runtime value, not a namespace |
| `userToken === process['env'].API_TOKEN` | object is a MemberExpression, not an Identifier |
| `userToken === secrets[API_TOKEN]` | computed — the property name is unknowable |

This **deletes** the `isProcessEnv` special case rather than extending it to bracket notation: `process` is lowercase, so neither spelling was ever a namespace. A narrower rule that needs fewer exceptions.

Measured cost on the corpus: 26 → 27 findings. The one regained is `this.auth !== this.ANY`, which has no Identifier object. Fixtures and the changeset are updated.

### ✅ Invalid regex syntax reported as ReDoS — **fixed, and the layer is gone**

Right, and following it through made the conclusion bigger than the fix. Parse failure is now terminal — `new RegExp("(a+")` throws at construction and can never backtrack, so it gets no report.

Once invalid syntax is separated out, the heuristic layer had nothing left to do: a pattern either fails to parse, or reaches the analyser and gets a verdict. Its only remaining effect was overruling clean verdicts. So it is **deleted** — the `REDOS_PATTERNS` table, `hasReDoSVulnerability`, `generateFixSuggestions`, and the four suggestion messageIds.

`allowCommonPatterns` is accepted-and-ignored rather than removed, so configs that set it keep loading; it is marked deprecated and goes in the next major.

Rather than replace the synthetic fallback fixtures, they are deleted with the layer. Coverage is back to 100% through the single reachable catch.

### ✅ `AST_NODE_TYPES` instead of string literals — **fixed** in both call sites.

### ❌ Split benchmark execution from publication — **declined here, worth its own issue**

Correct as a design point, but out of scope for this PR. My change to `weekly-benchmark.yml` only **reduced** permissions — top-level `contents/pages/id-token: write` moved down onto the single job that needs them, addressing a Scorecard alert. Splitting measurement from publication into two jobs with an artifact handoff and a PR-based commit path is a workflow redesign that touches Supabase and Pages wiring, and it does not belong in a PR about rule false positives.

Filing it separately is the right move; happy to take it next.

PR #537, opened 2026-08-12
fix(workspace): close OpenSSF Scorecard vulnerability, pinning and SAST findings

ofri-peretz:
Closes the OpenSSF Scorecard alerts that are fixable without a second maintainer.

## Vulnerabilities: 0 → clean (`high`, alert 1193)

Scorecard reported **29 OSV vulnerabilities**; root `npm audit` reported **one**. The whole gap was two vestigial package-level lockfiles in `eslint-plugin-browser-security` and `eslint-plugin-secure-coding`, pinning `ajv@6.12.6` and `brace-expansion@1.1.12/2.0.2` — one ReDoS, three DoS.

They were unreachable by any install path (measured, npm 11.16.0 / Node 24):

| path | result |
|---|---|
| `npm ci` inside the package dir | resolves against the **root** lockfile → fixed `ajv@6.15.0` / `brace-expansion@2.1.4` |
| `npm ci --workspaces=false` | **fails** — pins `@interlace/eslint-devkit@1.2.1` against a current `1.13.0` |
| copied outside the workspace | **fails identically** |

`package-lock.json` is excluded from published tarballs, so consumers never saw them either. The impact was on scanners, not installs: OSV reads lockfiles off disk regardless of npm's resolution rules.

## Pinned-Dependencies: 3 of 4 closed

- **`benchmark.yml`** (1322, 1323) — the last two tag-referenced actions pinned to the SHAs the rest of the file already uses.
- **`weekly-benchmark.yml`** (1317) — `npm install` → `npm ci` against its committed workspace lockfile.
- **`Dockerfile`** (1306) — exact versions on an `npm install` line pin the *direct* deps and nothing beneath them, so two builds could ship different code. Manifest + lockfile moved to `docker/image/`, installed with `npm ci`. **Verified** by building `linux/arm64` and `linux/amd64` and by the image's own in-container selftest; lockfile has no platform-constrained entries and audits clean.
- **`codecov.yml`** (1305) — `codecov-cli==11.3.1` pinned the CLI but let every transitive dep re-resolve from PyPI. Now `--require-hashes` against a committed set. **Verified** to install and run `codecovcli 11.3.1` on Python 3.12.

### Deliberately not pinned: `@lhci/cli` in `lighthouse.yml` (1304)

Pinning means committing a lockfile, and OSV reads committed lockfiles whether the tree is dev or prod. Measured: `@lhci/cli@0.14.0` resolves **9 advisories** (`tmp` high, `uuid` and `@lhci/cli` moderate, 6 low), with no release that avoids them.

That trades one **medium** Pinned-Dependencies alert for 9 findings on the higher-weighted **Vulnerabilities** check, while runtime exposure is identical — the global install runs the same code. Recorded in `SCORECARD-HARDENING-PLAN.md`.

## SAST: 25/30 → every commit (alert 1229)

This reverses the earlier recommendation in `SCORECARD-HARDENING-PLAN.md` ("don't — 0.06 points isn't worth a 4-minute job on every push").

Two things changed it. **Scope**: the trigger is `push: branches: [main]`, so it fires once per merge, not once per commit on a feature branch. **The gap was never only frequency**: CodeQL fired solely on `pull_request`, so anything reaching `main` outside that path — automation commits, changeset releases, PRs merged while still draft without `run-full-ci` — was never analysed at all.

The `concurrency` block had to change with it. Ref-keyed with `cancel-in-progress: true`, two merges landing close together would cancel the first commit's analysis and leave it permanently unscanned, recreating the gap. Push events are now keyed on `github.sha` and never cancelled; PR pushes still supersede each other.

## Not addressed (needs a second person or a policy call)

`CodeReviewID` (0/18 — self-merged), `Contributors`, `FuzzingID`, `CIIBestPracticesID`. `SecurityPolicyID` still flags despite `SECURITY.md` carrying the advisory URL and disclosure email — worth investigating rather than guessing at.

## Verification

`lefthook run pre-commit --all-files` green — scope-audit, stale-build-artifacts, oxlint, shim-drift, markdown, tests-affected (23/23).


ofri-peretz:
You're right, and the changeset was self-contradicting — it said the lockfiles "were never used for installs" and then claimed `npm ci` in those directories installed the vulnerable tree. Measured all three paths on npm 11.16.0 / Node 24:

- `npm ci` **from inside the package directory**: resolves against the root lockfile and installs the *fixed* `ajv@6.15.0` / `brace-expansion@2.1.4`.
- `npm ci --workspaces=false`: **fails** — `EUSAGE`, lock file's `@interlace/eslint-devkit@1.2.1` does not satisfy `@interlace/eslint-devkit@1.13.0`.
- Package copied outside the workspace: **fails identically**.

So there was no install path at all, not even the conditional one you asked me to state — the files are stale enough to be internally inconsistent. `package-lock.json` is also excluded from published tarballs, so consumers never saw them.

The changeset now says the impact was on scanners rather than installs: OSV reads lockfiles off disk regardless of npm's resolution rules, which is why these two files produced all 29 Scorecard vulnerabilities while root `npm audit` stayed clean. Removal is still correct, just for a different reason than originally written.

PR #542, opened 2026-08-12
fix(workspace): close two CodeQL findings and the last OSV advisory

ofri-peretz:
Three related fixes, all in the same failure class: **dependency state that only a scanner can see**.

The first two surfaced *because of* the CodeQL push-to-main trigger added in #537 — code that had never been analysed, since CodeQL previously only fired on `pull_request`.

## 1. Insecure temporary file — `js/insecure-temporary-file`, high, ×2

`scripts/corpus-scan.ts` built its scratch space at a fixed `os.tmpdir()/interlace-corpus-scan`, then wrote a `package.json`, ran an `npm install`, and generated an ESLint config inside it. On a multi-user machine that is a world-writable path an attacker can pre-create as a symlink and have us write through (CWE-377 / CWE-379). The script is documented for local use (`--update`), so this is not CI-only.

**`mkdtemp` is the usual answer and is wrong here.** The path is *deliberately* stable because it caches clones of eight large repositories between runs — a fresh directory per run discards exactly what it exists for.

Moved to the user cache directory (`XDG_CACHE_HOME`, else `~/.cache`): keeps the caching, removes the shared namespace. `mode: 0o700` for an unusual umask. The workflow doesn't cache the old path, so nothing downstream breaks.

## 2. esbuild `GHSA-g7r4-m6w7-qqqr` — the last advisory after #537

`apps/docs` resolved a nested `fumadocs-mdx@14.2.6` pulling `esbuild@0.27.7` (vulnerable range: introduced 0.27.3, fixed 0.28.1). Root already had `0.28.2`.

**Nothing required 14.x.** Root pins `fumadocs-mdx@15.2.2`; `apps/docs` declares `^15.2.2`. npm reported the nested copy `invalid` against apps/docs' own range — and reproduced it on every install anyway, because `npm install` does not prune stale entries. So apps/docs was running a **major version behind its own manifest**.

## 3. The guard, and the three violations it found

`scripts/__tests__/orphan-lockfile-lock.test.ts`. This class is invisible to everything already running: `npm audit` cannot see a lockfile npm never reads, `npm install` reproduces stale entries rather than pruning them, and nothing errors.

Two assertions, both **verified to fail on a broken tree** before being trusted:

1. **No `package-lock.json` at a workspace member root.** Scoped to `<member>/package-lock.json` exactly — lockfiles *deeper* inside a member (`packages/x/benchmark/`, `benchmarks/suites/*/`) are standalone rigs that get installed on their own, so they're maintained. Only one beside a member's own `package.json` is unreachable by npm.
2. **No nested entry violating its own member's declared range.** Nesting per se is legitimate (it's how npm resolves conflicting majors); a nested version failing the member's *own* range is not.

The second found three more, all older than what the member declares:

| member | package | locked | declared |
|---|---|---|---|
| `apps/docs` | `@types/node` | 22.19.7 | `^26.2.0` |
| `packages/eslint-formatter` | `eslint` | **8.57.1** | `^10.8.1` |
| `benchmarks` | `eslint-plugin-unicorn` | 62.0.0 | `^73.0.0` |

The formatter one is the notable one: **built and tested against ESLint 8 while declaring 10**, in a repo whose product targets ESLint 10. Its peer range still spans `^8.40 || ^9 || ^10`, so the compat matrix is unaffected — this only corrects what the lockfile installs.

## Verification

- `npm audit` → **0 vulnerabilities** (was 1)
- `turbo run build` 33/33, `turbo run test` 56/56
- `eslint-formatter` re-run with `--force` (the cached result predated the change): resolves **eslint 10.8.1**, 100% coverage
- `npm ci --dry-run` resolves against the rebased lockfile — two lockfile changes merged textually, so checked semantically rather than assumed
- `lefthook run pre-commit --all-files` green; pre-push typecheck + lockfile + build-and-test green

No package changes, so no changeset.


ofri-peretz:
Both findings were right; fixed in `4b4e2ef4`.

**On `XDG_CACHE_HOME` and `mode: 0o700`** — agreed, and the mode was doing even less than flagged: it applies *only when creating*, so on an existing directory it is a no-op. A pre-created world-writable directory or symlink was still accepted.

`scripts/lib/private-cache-dir.ts` now fails closed: ignores a relative `XDG_CACHE_HOME` (the XDG spec says a relative value is invalid and must be ignored — honouring one resolves the cache against the process cwd), refuses any path inside the *real* tmpdir, and `lstat`s every component from the cache root down, rejecting symlinks, foreign ownership, and group/world-writable modes.

It **refuses rather than repairs**. `chmod`-ing a directory we don't own, or unlinking a symlink we didn't create, is how a hardening step becomes its own vulnerability.

**On the missing lock** — added, though not in the suggested shape. That test recomputes `path.join(cacheHome, …)` and asserts the result doesn't contain `os.tmpdir()`, which passes regardless of what the script does: it tests its own arithmetic, not the source. `scripts/__tests__/private-cache-dir.test.ts` locks two separate things instead:

1. the **behaviour** of the guards against a real filesystem — symlink, `0o777`, file-instead-of-directory, relative path, tmpdir containment;
2. that `corpus-scan.ts` still **routes through** them — a bare `mkdirSync` reintroduced beside the call would bypass every check above.

The lock had to be a separate module: `corpus-scan.ts` calls `process.exit(main())` at module scope, so importing it from a test would run a full corpus scan.

**It caught a real bug in my first implementation.** `safeRealpath` fell back to `path.resolve` for a not-yet-existing path, so on macOS — where `os.tmpdir()` is `/var/folders/…`, a symlink to `/private/var/folders/…` — an unresolved candidate never string-matched the resolved tmpdir and the containment check silently passed. It now resolves the nearest existing ancestor and re-appends the missing tail.

16/16 across the three new lock files, and `ensurePrivateDir` verified against the real `~/.cache` path it has to work with in practice.

PR #570, opened 2026-08-16
chore(docs): refresh plugin-stats for the published versions, superseding #569

ofri-peretz:
## Summary

The docs homepage reads `plugin-stats.json` for each plugin's published version, and it still showed pre-release numbers. Regenerated against what npm now serves:

| Package | Version |
|---|---|
| `eslint-plugin-browser-security` | 1.4.1 |
| `eslint-plugin-node-security` | 4.13.1 |
| `eslint-plugin-secure-coding` | 4.3.0 |
| `eslint-plugin-jwt-security` | 3.0.1 |
| `eslint-plugin-express-security` | 3.1.1 |
| `eslint-plugin-react-features` | 1.3.0 |
| `eslint-plugin-react-a11y` | 2.2.0 |

## Why not just merge #569

#569 contains the same change and **cannot** merge. It was opened by `github-actions` using `GITHUB_TOKEN`, which by design does not trigger workflow runs — so its required checks never report and it sits `BLOCKED` with **zero** checks. Same reason #561 was closed earlier today.

## The pattern is now three-for-three today

1. **#561** — closed, unmergeable, work redone by hand
2. **#564** (release) — needed manual approval of its runs **twice**, because each force-push re-armed `action_required`
3. **#569** — this one

Fixing the token on the workflows that open these PRs (a PAT or GitHub App token instead of `GITHUB_TOKEN`) ends all three. #568 only stopped the review job going red on bot PRs; it does not make their checks run.

## Test plan

- [x] `turbo run test --filter=docs` — 20/20
- [x] Diff is exactly the eight version strings; rule counts unchanged

## Summary by CodeRabbit

* **Chores**
  * Updated plugin metadata to reflect the latest versions across security, React, accessibility, JWT, and Express plugins.
  * Refreshed the generated timestamp.
  * Plugin totals and counts remain unchanged.

### DIFF
--- a/apps/docs/src/data/plugin-stats.json
+++ b/apps/docs/src/data/plugin-stats.json
@@ -5,7 +5,7 @@
       "rules": 46,
       "description": "ESLint plugin for browser security — detects DOM XSS, postMessage abuse, tokens in localStorage, insecure cookies, clickjacking, mixed content, and CSP gaps",
       "category": "security",
-      "version": "1.4.0",
+      "version": "1.4.1",
       "published": true
     },
     {
@@ -13,7 +13,7 @@
       "rules": 42,
       "description": "ESLint plugin for Node.js security — detects command injection, path traversal, SSRF, zip slip, and weak crypto (MD5/SHA-1, ECB, static IV) in fs, child_process, vm, and crypto",
       "category": "security",
-      "version": "4.13.0",
+      "version": "4.13.1",
       "published": true
     },
     {
@@ -21,7 +21,7 @@
       "rules": 33,
       "description": "ESLint plugin for secure coding — detects LDAP, XPath, XXE, GraphQL and template injection, unsafe deserialization, ReDoS, missing authentication, and PII in logs",
       "category": "security",
-      "version": "4.2.0",
+      "version": "4.3.0",
       "published": true
     },
     {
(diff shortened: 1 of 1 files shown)


ofri-peretz:
Re: regenerating `interlace-numbers.json` so its `generatedAt` matches — declining, because the mismatch is the designed behaviour and it is pinned by a test.

`apps/docs/scripts/lib/write-json-if-changed.ts` deliberately skips the write when only the timestamp moved, and `apps/docs/src/__tests__/generated-data-write-lock.test.ts` locks it with two cases:

- `does not touch the file when only a top-level timestamp moved`
- `leaves the old timestamp in the file when it skips the write`

So a differing `generatedAt` is the guard working, not drift. Forcing a rewrite would fail that lock and put a timestamp-only diff in every stats sync — which is the churn the helper exists to prevent.

The finding also notes the substance is fine, and I verified it independently: `interlace-numbers.json` has `rules.total: 477` / `plugins.total: 30`, and `plugin-stats.json` has `totalRules: 477` / `totalPlugins: 30`. The numbers agree; only the date differs, because that file genuinely has not changed since the 13th.

No aggregate assertions added either — `numbers-manifest-lock.test.ts` already covers the totals.

PR #572, opened 2026-08-16
ci(workspace): make the docs-data sync PR actually mergeable, and self-merging

ofri-peretz:
## The actual problem

The sync PR **could never merge**. It is opened with `github.token`, and GitHub does not fire `on: pull_request` workflows for a PR created with `GITHUB_TOKEN` — so its required contexts never report and it sits `BLOCKED` with **zero** checks indefinitely.

#561 and #569 both died this way today. Both had to be regenerated by hand on a human branch and closed.

That is the same failure the existing comment in this file describes (`git push` to main → `GH006` → data thrown away), just one step later: the sync succeeds and its output is discarded anyway.

## This is not the bypass token that comment rejects

> *"A bot token that bypasses protection would clear the red X by punching a hole in the rule instead."*

Agreed, and nothing about branch protection changes here. The PR still runs every required check and still has to go green. The PAT only makes those checks **run**. `--auto` merges it once they pass.

`changesets-pr.yml` already uses this exact pattern.

## Also

- Auto-merge is re-asserted on **every** run, not just at creation — a force-push resets the PR's checks and disarms it.
- Failure to set auto-merge is a `::warning::`, not an error; the PR still exists and merges by hand.

## On the trigger

Worth saying explicitly: it is **not** a daily schedule. It is `push to main on packages/**`, so it fires when a plugin actually changes. It felt daily because today was a release day.

## Test plan

- [x] `npm run lint:workflows` — 34 files pass
- [x] YAML parses
- [x] Token falls through `RELEASE_BOT_PAT → AGENTS_REPO_PAT → github.token`, so it degrades to today's behaviour if no PAT resolves

## Summary by CodeRabbit

* **Chores**
  * Improved automated documentation synchronization checks for generated pull requests.
  * Re-enabled automatic squash merging after successful synchronization.
  * Added graceful handling when automatic merging cannot be enabled.

### DIFF
--- a/.github/workflows/docs-data.yml
+++ b/.github/workflows/docs-data.yml
@@ -46,6 +46,16 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           fetch-depth: 1
+          # The push below has to use the same token the PR is opened with.
+          # `actions/checkout` persists whatever it is given into the local git
+          # config, defaulting to `github.token` — so setting GH_TOKEN for the
+          # `gh` CLI alone is not enough: `git push` would still authenticate as
+          # GITHUB_TOKEN, the resulting `synchronize` event would fire no
+          # workflows, and the PR's checks would go stale on every run after the
+          # first. `--auto` then has nothing to merge on. Needs `contents: write`
+          # on whichever PAT resolves; without one this falls back to
+          # `github.token` and behaves exactly as it does today.
+          token: ${{ secrets.RELEASE_BOT_PAT || secrets.AGENTS_REPO_PAT || github.token }}
       - uses: ./.github/actions/setup
         with:
           turbo-cache: "false"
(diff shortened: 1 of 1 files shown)


ofri-peretz:
Both addressed in `16e45714`.

**Use the selected token for the branch push** — correct, and it was a hole in this PR's own fix. `GH_TOKEN` covers the `gh` CLI, but `actions/checkout` persists its own token and defaults to `github.token`, so `gh pr create` used the PAT while `git push` did not. The first PR would have triggered workflows and every force-push after it would have produced a `synchronize` event that fired nothing — checks going stale with `--auto` unable to merge, which is exactly the failure this PR set out to remove, one layer down.

The token now goes to `actions/checkout` as well, so the push and the `git ls-remote` lease read authenticate with it:

```yaml
token: ${{ secrets.RELEASE_BOT_PAT || secrets.AGENTS_REPO_PAT || github.token }}
```

It needs `contents: write` on whichever PAT resolves. With none set the expression falls back to `github.token` and the workflow behaves exactly as it does today, so this cannot regress anything. Comment updated to say the PAT is conditional.

**Resolve exactly one `main` PR before enabling auto-merge** — also correct. Lookup is now `--base main --limit 1` with `.[0].number // empty`, extracted into a `find_pr` helper used by both the existing-PR and just-created paths. An empty result is a hard `::error::` and `exit 1` rather than falling through to `gh pr merge ""`, where the `|| echo` would have converted it into a warning and reported a successful sync that enabled nothing.

PR #584, opened 2026-08-20
fix(node-security): an eval flag only means what the binary parsing it says

ofri-peretz:
## A security rule committing the defect it polices

`usesShell` treated any of `-c`, `-e`, `/c` in argv as proof that the next entry is source text — whatever program was being run.

Found on the 20-repository real-source corpus, in n8n's `scripts/dev-up.mjs`:

```js
execFileSync('gh', ['codespace', 'ports', 'visibility', `${port}:org`, '-c', name], { stdio: 'ignore' });
```

`-c` there is **gh's own `--codespace`**. Every argv entry is a literal or a template of literals, no shell is anywhere near it, and the rule reported **CWE-78 command injection at CVSS 9.8**.

The comment defending the old behaviour said reading the flag *"rather than enumerating interpreters"* avoided a list to maintain. That trades a list for a wrong answer — and deciding by a **token** instead of by the program that parses it is exactly what `lint:name-inference` exists to catch.

## The gate

`EVAL_CAPABLE_BINARIES` — every shell, plus `node`, `deno`, `bun`, `python`, `ruby`, `perl`, `php`, `osascript`.

It is deliberately one-sided: it **only ever suppresses**, and only when the command is a **literal** naming a binary we can place. A command we cannot name keeps the old conservative reading, because then we genuinely don't know what parses the flag:

```js
execFileSync(bin, ['-c', name])   // still reports — bin may be a shell
```

## Evidence

| | before | after |
|---|---|---|
| real-source findings (21,394 files, 3.10M lines) | **7** | **5** |
| `sh -c` / `node -e` / `cmd /c` | report | report |
| package suite | — | 2694 passing, 100% statements + branches |

FN guards for all three eval spellings plus the unnameable-binary case are pinned in the new test file, and the lock fails on the unfixed rule.

---

This is the second half of the `effectiveFp` work on this rule. [#581](https://github.com/ofri-peretz/eslint/pull/581) fixed what a finding *says*; this fixes a finding that should not exist. Also included: the computed-key recall fix, already reviewed in [#583](https://github.com/ofri-peretz/eslint/pull/583) and re-pushed here after that branch was deleted on merge.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of filesystem access paths that use unresolved computed keys, while preserving valid static-key handling.
  * Refined child-process checks so eval-style flags are reported only for shell and interpreter binaries, reducing false positives for other commands.
* **Tests**
  * Added coverage for computed filesystem paths and supported shell/interpreter invocation patterns.
* **Documentation**
  * Updated plugin metadata and release information across security-focused plugins.


ofri-peretz:
Confirmed and fixed in `43438e2` — you were right, and it was the same defect one level up from the one this PR set out to fix.

Gating on *"is this an interpreter"* stopped `gh … -c NAME`, but kept one flag set shared across every runtime. A flag means only what the program parsing it says it means, and that holds *between* interpreters just as much as between an interpreter and `gh`:

| | evaluates | does not |
|---|---|---|
| `php` | `-r` | `-e` — that is `--profile-info` |
| `deno` | `eval` (subcommand) | `-e` — no such option |

Both of your cases now behave, in both directions. Checking `--help` while fixing them turned up two more the original list never had: **`node -p`** evaluates and prints, and **`perl -E`** is `-e` with feature bundles enabled — both were false negatives before this PR, not just before your comment.

`EVAL_INVOCATIONS` now maps each binary to its own tokens. Sources: `--help` on this machine for node, python3, ruby and perl; the CLI manuals for php and deno. Shells are deliberately absent from the table — a literal `sh`/`bash`/`cmd`/`powershell` command is already decided by `SHELL_BINARIES` before flags are read. A binary we cannot name still gets the conservative union, which is now complete rather than partial.

13 cases pinned, both directions for php/deno/python plus every regression guard. 2708 passing, 100% statements and branches.

PR #609, opened 2026-08-22
fix(maintainability): identical-functions stops reporting on code nobody edits

ofri-peretz:
## Summary

`identical-functions` produced **3,707** findings on the pinned 8-repository corpus. Two classes were nine tenths of it — both correct detections of real duplication, and neither actionable. A correct finding nobody acts on counts against the rule exactly as a wrong one does.

| | findings |
|---|---:|
| before | 3,707 |
| after the test-file fix | 1,676 |
| after the generated-file skip | **371** |

### Test files — 52.9%

`ignoreTestFiles` already defaulted to `true`. Its predicate was `/\.(test|spec)\.[jt]sx?$/` — a **basename** test. okta-auth-js keeps its entire suite under `test/spec/**` with plain filenames (`test/spec/TokenManager/browser.ts`), so on that repo the default-on option did nothing at all.

Now uses the devkit's shared `isTestFilePath`: basename for `*.test.*`, exact path **segment** equality for the directory case. Segment equality is the point — `filename.includes('test')` makes the same file report from `~/src/app` and stay silent from `~/latest/app`, which has shipped here once already.

A repeated `it()` block is not a maintenance defect. The repetition *is* the test: each case restates its setup so it reads independently when it fails.

### Generated files — 35.6%

New `ignoreGeneratedFiles`, default `true`. twilio-node's `src/rest/**` is emitted by OpenAPI Generator; every resource class gets the same `fetch`, `remove`, `page`. The duplication is real, and the header says "Do not edit the class manually".

Decided by the file's **own declaration** — an `@generated` docblock or a "this code was generated by" banner in the *leading* comments. Specifically not:

- **not the path.** `src/generated-reports/summary.ts` is a name, not a declaration, and still reports.
- **not anywhere in the file.** A generation phrase below the first statement does not silence the module — otherwise any comment buys a whole-file suppression for free.
- **not a bare "do not edit".** Hand-written files carry that too.

### The false-negative direction

Of the 507 files the generated skip silenced, **all 507** carry a banner in their first 2KB, and **all 507** are in twilio-node. The other seven repositories lost nothing. twilio-node keeps 15 findings, in its hand-written `src/base`, `src/jwt` and `examples`.

## Test plan

- [x] `maintainability`: 399 passing, **100%** statements / branches / functions / lines.
- [x] `eslint-devkit`: 1,743 passing, **100%** statements / branches / functions / lines.
- [x] New locks in `identical-functions-scope.test.ts` — **5 of the 10 fail on the unfixed rule**, verified by restoring `HEAD`'s version and re-running. The other 5 are the positive control and the guards that must hold either way.
- [x] FN guards both directions: `ignoreTestFiles: false` and `ignoreGeneratedFiles: false` each restore the finding.
- [x] `npm run lint:name-inference` — 314 rules scanned, no new site.
- [x] **CWE recall gate re-run in the same session**: TP=69 FP=0 FN=0, recall 100%. No CWE lost a detection.

## Note on the mock

`createWithMockContext` gained `getAllComments` and `getFirstToken`. A real `SourceCode` has both, and without them a rule reading the leading comment block throws inside a synthetic-AST test. The tempting alternative — making `isGeneratedFile` defensive about a shape ESLint always provides — would add a branch no real input can take, which in a repo held at 100% coverage is a permanent hole.

## After merge

The corpus budget records **published** behaviour, so it does not move until release. `--local` already shows the new number.


ofri-peretz:
## Second commit: the same predicate, two more rules

`createRule` gains an opt-in **`skipGeneratedFiles`**, the counterpart to the existing `skipTestFiles`, now set on `cognitive-complexity` and `max-parameters`.

The two flags are deliberately **not** symmetric, and the rule-creator already documents why: `skipTestFiles` stays off for these rules because a convoluted test is still a readability cost to whoever debugs it. Generated code has no such reader.

I measured the share **before** writing the change, then checked the change against the prediction:

| rule | predicted generated share | before | after | actual |
|---|---:|---:|---:|---:|
| `cognitive-complexity` | 32% | 1,409 | 965 | −32% |
| `max-parameters` | 27% | 407 | 296 | −27% |

Exact match both times, which is the evidence that the predicate silenced the generated set and nothing else — not a broader class that happened to land near the same number.

### Locks

- Devkit gate test with the three-way control: suppresses when opted in, **still reports in hand-written files** (else it passes on a rule that suppresses everywhere), and **CONTROL: reports when NOT opted in** (else it passes on a rule that never had the gate).
- Per-rule lock with a positive control each. Deleting `skipGeneratedFiles: true` from either rule turns **2 of the 6 red** — verified by doing it.

### Not for security rules

Documented on the option and in the gate test: generated code ships and runs, so an injection in it is a live vulnerability whoever typed it. This flag is for maintainability advice only.

- [x] `maintainability` 405 passing, `eslint-devkit` 1,747 passing, both **100%** on all four coverage axes.
- [x] CWE recall gate re-run after the devkit change: TP=69 FP=0 FN=0.
- [x] `lint:name-inference`: 314 rules, no new site.

PR #650, opened 2026-08-23
feat(scripts): list every PR in a release, in the release notes

ofri-peretz:
## Summary

Release notes carried the changeset text and nothing else. That says what changed
in our words, which is the right thing to lead with — but it cannot answer the
question a consumer actually arrives with: **which pull requests are in the version
I just installed?**

That is the question asked by someone whose reported false positive stopped
appearing, and by anyone deciding whether an upgrade is worth the churn. This week
made it concrete: nine rule fixes shipped across two PRs, and nothing in the
release notes told a reader which.

`scripts/prs-since-release.ts` finds the package's **own** previous tag — the
repo's tags interleave every package, so a repo-wide "last release" is the wrong
boundary — and lists the squash-merge PRs that touched it since.

- Tags sort by `-v:refname`, so `5.1.10` ranks above `5.1.9`. A lexical sort puts
  `5.1.9` last and silently truncates the range to a single release, which reads
  as a quiet changelog rather than a broken script.
- Reads git, not the GitHub API: tags and subjects are already in the checkout, and
  an API call would need a token and a rate-limit budget in a matrix job that runs
  once per package.
- Skips the changesets bot's own version-packages commit, which carries the bump
  and CHANGELOG for this very release.
- Emits nothing when there are no PRs, rather than a heading with an empty list.

## Test plan

- [x] Verified against all three packages released 2026-08-23; each lists exactly
      the PRs that touched it since its previous tag
- [x] `node scripts/lint-workflows.ts` — 34 workflows pass
- [x] `oxlint` and `prettier` clean

## Not in scope

The tag race that stopped every GitHub Release on 2026-08-23 — every package
published to npm, then `gh release create` failed with "tag exists locally but has
not been pushed". #647 already fixed that with `--target`.

## Summary by CodeRabbit

* **New Features**
  * Release notes now include pull requests affecting each package since its previous release.
  * Pull-request references are automatically grouped by package and linked for easier access.
  * Release notes remain concise when no relevant pull requests are found.

* **Documentation**
  * Added release-note guidance describing how package-specific changes are identified and presented.

### DIFF
--- a/.changeset/release-notes-list-prs.md
+++ b/.changeset/release-notes-list-prs.md
@@ -1,0 +1,22 @@
+---
+'eslint-plugin-node-security': patch
+---
+
+Release notes now list every pull request in the release.
+
+The changeset text says what changed in our words, which is the right thing to
+lead with. It does not answer the question a consumer actually arrives with —
+_which_ PRs are in the version I just installed — and that is the question asked
+by someone whose reported false positive stopped appearing, or by someone
+deciding whether an upgrade is worth it.
+
+`scripts/prs-since-release.ts` reads the package's own tags, walks the commits
+that touched that package since its previous release, and lists the squash-merge
+PRs. Per-package, because the repo's tags interleave every package and the
+question is always about one of them. Sorted by `-v:refname` so 5.1.10 ranks above
+5.1.9 rather than lexically below it, which would silently truncate the range to a
+single release.
+
+It reads git rather than the GitHub API: the tags and subjects are already in the
+checkout, and an API call would need a token and a rate-limit budget inside a
+matrix job that runs once per package.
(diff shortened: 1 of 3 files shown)


ofri-peretz:
Both CodeRabbit findings were real and are fixed.

**Fail when git fails** — the helper swallowed every error into an empty string, so a broken `git log` produced an empty PR list and exited 0. That is release notes silently omitting the one thing they exist to show, and it is the same shape as a lint harness reporting zero findings because it was wired to zero rules: the wrong answer is indistinguishable from the good one. It now prints the failing command and exits 1.

**Tag format** — the sharper catch. `release.yml` strips to the last slash, so `@interlace/eslint-devkit` is tagged `eslint-devkit@1.9.0`. I was searching the full package name, which matched nothing for **every scoped package** — and because "no previous tag" legitimately means "first release", it would have quietly listed the entire history rather than erroring. Verified: `eslint-devkit` now resolves to 1.16.1.

Worth noting that a script whose whole point is release-note accuracy shipped with a silent-failure path and a bug affecting every scoped package. Caught before merge, which is the system working.

PR #661, opened 2026-08-23
fix(docs): keyboard access for scrollable homepage code blocks

ofri-peretz:
## Summary

axe `scrollable-region-focusable` (WCAG 2.1.1, Keyboard, **serious**) fires on the docs homepage in both light and dark themes. It is two distinct regressions that surface at *different viewports*, so neither one alone reproduces the full picture:

| Element | Viewport | Overflow | Fix |
|---|---|---|---|
| Hero code block — standalone `<pre>` | ~390–500px | 562px content in a 466px box | `tabIndex={0}` + `role="region"` + `aria-label` |
| "What it catches" card snippets — `<pre>` inside the card `<Link>` | desktop | 320px content in a 204px box | `whitespace-pre-wrap break-words` (remove the scroll) |

The two fixes differ on purpose. The hero block is standalone, so making it focusable is correct — horizontal scrolling suits a config sample. The card snippets sit **inside** the card `<Link>`, so adding `tabIndex` there would nest a tab stop inside an anchor; removing the scrollable region entirely is the right fix for three short snippets.

## Test plan

- [x] Reproduced with axe-core 4.10.2 against production at two viewports — 3 nodes at desktop, 1 node at 390–500px width.
- [x] New lock: `apps/docs/src/__tests__/homepage-scroll-a11y-lock.test.tsx` (11 assertions). It locks the **general invariant** — every `overflow-x-auto` `<pre>` is focusable and named; card snippets wrap and carry no tab stop — not the specific literals, so it also catches a *new* scrollable block added later.
- [x] Assertions run against comment-stripped source via the audited `blankComments` helper from `lock-integrity.test.ts`, so they cannot be satisfied by a code comment.
- [x] **Mutation-verified:** reverting the hero fix turns the lock red (2 failures); reverting the card fix turns it red (1 failure); restoring both returns 11/11 green.
- [x] Full docs suite run from `apps/docs`: **82 files, 1680 passed, 0 failed**.
- [ ] Local `build-and-test` / `typecheck` pre-push hooks were bypassed — they fail only because this git worktree symlinks `node_modules` (Turbopack rejects the symlink; the same symlink resolves `@interlace/eslint-devkit` to a working copy with unrelated WIP). Relying on the required CI checks **TypeScript** and **Build (apps/docs)** for that coverage.

## After merge

Auto-deploy fires `deploy-docs.yml`. Re-run axe against https://eslint.interlace.tools at 390px and desktop to confirm zero `scrollable-region-focusable` nodes in both themes.

Note: the card snippets will now wrap onto a second line instead of scrolling — an intentional, visible change to the "What it catches" cards.


ofri-peretz:
Addressed the lock finding in `2ad2fe899` — the class-based fingerprint was a real vacuity risk, not just a maintainability note.

**Was:**
```tsx
const cardPres = preTags.filter((t) => t.includes('bg-fd-background/80'));
```
Restyle `CatchCard` and `cardPres` becomes `[]` — the loop body never runs, and every assertion inside it is silently skipped while a re-introduced `overflow-x-auto` ships.

**Now** the invariant is asserted by semantic position rather than styling, which is what we actually mean:

```tsx
const linkBlocks = code.match(/<Link\b[\s\S]*?<\/Link>/g) ?? [];
const presInLinks = linkBlocks.flatMap((b) => b.match(/<pre\b[^>]*>/g) ?? []);
expect(presInLinks.length).toBeGreaterThan(0); // non-vacuity guard
```

This also closes the gap you flagged at the end — a future card `<pre>` that is scrollable *but unlabeled* now fails, because any `<pre>` inside a `<Link>` is rejected for carrying `overflow-x-auto` at all.

**Mutation-verified against your exact scenario:** restyling to `bg-fd-card/80` *and* re-introducing `overflow-x-auto` turns the lock red. The old assertion passed that case.

On your parenthetical re-derivation: your correction is right. Reverting the hero fix removes `tabIndex={0}` and keeps `overflow-x-auto`, so `scrollable.length > 0` holds and the two `tabIndex`/`role` assertions fail — 2 red, as the PR body states.

Full docs suite after the change and a `main` merge: **83 files, 1688 passed, 0 failed.**

Note for the record: CodeRabbit hit its OSS rate limit on this PR and has not reviewed it.

PR #672, opened 2026-08-24
fix(release): lead release notes with what reaches consumers

ofri-peretz:
## Summary

Found by reading the first rollup the new system actually published ([release-2026-08-23-634bee07](https://github.com/ofri-peretz/eslint/releases/tag/release-2026-08-23-634bee07)):

> ✅ **Safe to upgrade.** No breaking changes: existing configs keep working as-is.
>
> ### 💥 Breaking changes
> - categorised changelogs, a cross-package rollup, and versioned apps — `docs` *(internal — not published)*

The banner and the first heading contradict each other. The per-line `(internal)` marker resolves it — but only after the reader has read the line, and **a heading is read before its contents**. In that release four of five entries were internal app churn, so the one fix a consumer actually cared about was buried under three sections that could not affect them.

Consumer-facing entries now lead; internal ones collapse into a `<details>` at the end. Nothing is dropped — the record stays complete, under a heading that says what it is.

Re-rendered against that exact release, it goes from four sections of noise to:

```
✅ Safe to upgrade. …
Released: 2 packages, 1 app.

### 🐛 Fixes
- `no-http-urls` no longer reports test material or bare scheme strings. — browser-security (#666)

<details><summary>Internal changes — apps and unpublished packages</summary>
```

## The systemic half: verifying the artifact

Everything upstream of the publish is unit-tested — the formatter, the rollup, the normaliser, the changeset gate. **Nothing tested the artifact.** Between a green suite and a reader on npm sit a markdown slice, a shell heredoc, `gh release create`, and GitHub's renderer. Every defect found in this system so far — including this one — was found by opening the published page and reading it, not by running the tests.

`scripts/verify-release-notes.ts` is that reading, automated:

| Rule | Catches |
|---|---|
| RN001/RN002 | Empty body, or the `--fallback` stub (CHANGELOG and published version disagree) |
| RN003 | No upgrade verdict |
| RN004 | Missing install line, or one naming the wrong version |
| RN005 | A bullet leading with raw `[#123](…)` link plumbing |
| RN006 | A split code span (odd backtick count) |

**Run against the last six releases it found 12 real issues — all in pre-#660 notes — and passed both releases the new system produced.** It distinguishes old format from new on its own, which is the evidence that the format change was worth making.

Wired into `release.yml` against exactly the tags each run creates (not the latest N, which would re-report historical notes forever), and advisory: the packages are on npm by then, so a red X cannot un-publish anything and would only teach people to ignore red Xs.

## Test plan

- [x] `vitest run` across the six release-tooling suites — **187 pass**.
- [x] New `scripts/__tests__/verify-release-notes.test.ts` — 11 assertions, fixtures taken from the real `browser-security@2.0.4` body and the real legacy shape. Covers both failure modes: missing a bad note, and crying wolf on a good one.
- [x] Three new rollup tests: consumer sections lead, the internal section is omitted when empty, and the internal entry is still listed.
- [x] Verified by re-rendering the actual production release range.
- [x] oxlint, prettier, `lint-workflows`, `changeset:lint` clean.

## After merge

Every release run reports on its own notes in the job summary. Audit any range by hand with `npm run release:verify-notes -- --latest=10`.

## Summary by CodeRabbit

- **New Features**
  - Release notes now highlight consumer-facing changes first.
  - Internal changes are grouped in a collapsible section for clearer reading.
- **Documentation**
  - Added guidance for validating published release notes locally and in the release workflow.
- **Quality Improvements**
  - Added automated checks for upgrade guidance, installation commands, links, formatting, version consistency, and missing release content.
  - Improved handling of release notes with no internal changes.
- **Tests**
  - Expanded coverage for release-note organization and validation scenarios.


ofri-peretz:
## RN004 finding — fixed, and it was a good catch

You're right, and the failure mode is worse than "fires falsely": it would have fired on **every scoped package, every run**, which is exactly how a checker gets muted and then ignored.

`release.yml` builds the tag with `sub("^@[^/]+/"; "")`, so `@interlace/eslint-devkit@2.0.0` publishes as tag `eslint-devkit@2.0.0` while the body carries the full name. `body.includes(install)` fails on the `@interlace/` sitting between the prefix and the name.

RN004 now matches the version and the unscoped name with an **optional** scope prefix, anchored with `\b` so `@interlace/not-eslint-devkit` does not satisfy a check for `eslint-devkit`.

### The test was the real problem

My unit test paired a scoped tag with a scoped body — `verify('@interlace/eslint-devkit@2.0.0', …)`. That shape cannot occur, because `release.yml` never emits a scoped tag. So the test reported green on a check that was broken for every scoped release.

Replaced with the pairing that actually happens (unscoped tag, scoped body), plus two negative cases the old test could not have caught:

- a scoped package whose install line names a **stale version**
- a **different package** that merely ends with the same characters

13 assertions in that suite now, 45 across both affected suites.

PR #682, opened 2026-08-24
fix(docs): the changelog API returned "date": "Unknown" for every modern release

ofri-peretz:
## Summary

`/api/changelog` is a **documented public endpoint** — `advanced/changelog.mdx` publishes `curl` examples for it. Its parser reads the release date out of the version heading:

```js
const versionRegex = /^##\s*\[?(\d+\.\d+\.\d+)\]?\s*[-–(]?\s*(\d{4}-\d{2}-\d{2})?/gm;
const date = sections[i + 1] || 'Unknown';
```

That works for the legacy keep-a-changelog format. **Changesets writes `## 1.4.1` with no date at all** — so from the day this repo adopted changesets, every entry the endpoint returned carried `"date": "Unknown"`. A documented public field that has never held a date for a modern release.

The real date is the release's git tag, which `sync-changelog.ts` already resolves for all 458 releases. The route now consults that first, falls back to the heading date, and only then admits `"Unknown"`.

**Measured on `eslint-plugin-jwt-security`:** 8 of 10 entries gain a real date. The two that remain unknown genuinely have no tag.

`type` gets the same treatment. It was a substring guess — `content.includes('fix')` fires on any entry whose prose happens to contain the word — where the changeset's own conventional-commit prefix records what the author *declared*. The guess stays as the fallback for entries with no declared kind.

**Response shape is unchanged**; only the values improve. This matters because it's a public contract.

## Test plan

- [x] `vitest run` in `apps/docs` — **94 files, 1990 tests pass**.
- [x] `next build` clean; served the built app and called the endpoint:
  - before: `with real date: 0 / 10`, types from keyword guessing
  - after: `with real date: 8 / 10`, types `{security: 5, feature: 4, breaking: 1}`
- [x] Typecheck clean; prettier and markdownlint clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved changelog release date accuracy using available release metadata, with sensible fallbacks when dates are unavailable.
  - Improved release type classification by prioritizing declared release types and retaining fallback detection for older entries.
  - Maintained the existing changelog API response format for compatibility.


ofri-peretz:
Fixed — and the finding is sharper than "the contract says tag first".

`sync-changelog.ts` builds the very data this route reads, and it already uses `tagged ?? headingDate`. My route had the reverse, so the two disagreed about the same field: for any legacy release carrying both a tag and a hand-written heading date, the site and the API would report different dates for the same version.

On the substance you're right too. The tag is when the release actually published; a heading date was typed by hand and records when someone edited the changelog, which is often a different day. Where both exist and differ, the tag wins.

Now `facts?.date || sections[i + 1] || 'Unknown'`, with a comment naming the sync script so the two stay tied together.

PR #703, opened 2026-08-24
feat(docs): DS-grounded UX pass — scorecard, articles, changelog, playground

ofri-peretz:
## Summary

The experience pass over the site's tabs, implemented with `@interlace/ui` primitives and the documented philosophies (CTA, layout, landscape framing). Every change either uses an existing DS affordance or fixes one at its source.

### Scorecard — value before receipts, peers not competitors
- **`STACK_LABELS` said "Competitor (ESLint)"** on two public surfaces (the `/scorecard` page and the generated markdown reports) — a violation of the landscape framing rule. Now **Peer**; internal type keys keep their historical names. Locked by `landscape-framing-scorecard.lock.test.ts`, scoped to reader-visible values only.
- The page opened with **Provenance** — version numbers and a JSON path before any value statement. A new `ScoreSummary` strip leads instead: head-to-head warm wins, green-field count, and an honest **snapshot-age chip** — every figure derived server-side from the snapshot, never hand-typed. Provenance moves to the end, where receipts belong.

### Articles — one name, no zero-view chips
- One page, three names (nav *Articles* / title *Technical Articles* / h1 *Technical Insights*) → **Articles** everywhere.
- `@interlace/ui` **ArticleCard rendered "👁 0" on every card** (dev.to's public API reports zero views). The DS block now renders nothing for 0 — absence is the honest presentation of absence. Changeset included (`@interlace/ui` patch).

### Changelog — the cloud collapses, the constraints survive
- 34 filter chips wrapped into ~5 desktop rows before the first release. The cloud now lives in a **native `<details>`** — zero JS, links stay in the DOM for crawlers, keyboard-operable natively — with an always-visible row carrying only the reset chip and the active filter. The original design's constraints (URL state, no hydration, `aria-current`) are all preserved. The page lock evolved to pin the new invariant with JSX-shaped anchors (immune to the docstring trap).

### Playground + Getting Started
- `SectionHeader as="h1"` — the page had no h1; the DS block already supported the prop.
- The copy button uppercase-transformed a filename (`ESLINT.CONFIG.JS` is a wrong name, not a style).
- **Getting Started gets install-in-30-seconds above the fold** — the same inversion shipped for plugin overviews, with the instrumented `InstallSnippet`.

## Deliberately deferred

Docs-sidebar de-duplication (the 8 mirrored nav links): needs fumadocs `on:'nav'` semantics verified against the **deployed** 16.14.5 (local resolves 16.11.5), and a wrong guess silently breaks mobile navigation.

## Test plan

- [x] Full docs suite: **89 files, 1800 passed, 0 failed** — including the evolved changelog lock, the new framing lock, and all pre-existing scorecard/stats/articles locks.
- [x] New framing lock is comment-trap-proof (JSX/value-scoped anchors) and constrains only reader-visible strings.
- [x] All summary figures on the scorecard derive from the loaded snapshot at render time — nothing hand-typed.

### DIFF
--- a/.changeset/ds-article-card-zero-views.md
+++ b/.changeset/ds-article-card-zero-views.md
@@ -1,0 +1,9 @@
+---
+'@interlace/ui': patch
+---
+
+fix: ArticleCard renders no views chip for zero views
+
+Sources that cannot report views (dev.to's public API) return 0, and the card
+displayed a literal "👁 0" on every article. A zero-view chip reads as product
+failure, not information — absence is the honest presentation of absence.
--- a/apps/docs/content/docs/getting-started/index.mdx
+++ b/apps/docs/content/docs/getting-started/index.mdx
@@ -14,6 +14,25 @@ import {
   CheckCircle,
 } from 'lucide-react';
 
+## Install in 30 seconds
+
+The most-adopted starting point — general secure-coding rules that work in any
+JavaScript or TypeScript project:
+
+<InstallSnippet packages="eslint-plugin-secure-coding" dev />
+
+Then enable the recommended preset and lint:
+
+```js
+// eslint.config.js
+import secureCoding from 'eslint-plugin-secure-coding';
+
+export default [secureCoding.configs.recommended];
+```
+
+Framework-specific plugins (Express, React, MongoDB, AI SDKs, …) are one
+[catalog](/plugins) away — every plugin installs the same way.
+
 ## What is ESLint Interlace?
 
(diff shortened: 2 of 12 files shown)


ofri-peretz:
Consolidated into #700 — the two branches were racing each other for the same merge window against a fast-moving main. All 12 files (scorecard reframe + ScoreSummary, articles naming + zero-views DS fix, changelog disclosure, playground h1/casing, getting-started install, framing lock, @interlace/ui changeset) are now on feat/wide-viewport-smoke-gate, suite green at 89 files / 1816 passed.

PR #727, opened 2026-08-25
fix(node-security): a bare .extract() is not archive extraction

ofri-peretz:
## Summary

`no-zip-slip` matched `.extract()` on any receiver. Two shapes found by scanning outreach candidates, neither of which touches an archive:

```js
// passbolt/passbolt_styleguide — src/shared/models/entity/user/usersCollection.js:75
const uniqueIdsSetCache = new Set(this.extract("id"));

// nioc/node-red-contrib-opentelemetry — lib/opentelemetry-node.js:259
return propagator.extract(context.active(), msg.req.headers, defaultTextMapGetter)
```

**22 findings** on passbolt — an open-source password manager — and **3** on node-red-contrib-opentelemetry. Neither file contains the substring `zip`, `tar` or `archive` anywhere. Both were about to become the entire content of an outreach pull request.

## Why the archive-context guard did not catch it

The guard was circular:

```js
if (isArchiveExtraction(node) || namesArchive(node.callee)) hasArchiveContext = true;
```

`isArchiveExtraction()` matched a bare `.extract()`, so it established the context that `isArchiveExtraction()` then relied on — the call being judged was its own evidence that the file handles archives.

Two changes:

- Context comes only from `namesArchive`, never from the extraction test itself.
- The four verbs that are ordinary English — `extract`, `extractAll`, `unzip`, `untar` — require a receiver that names an archive. `extractAllTo` and `extractArchive` belong to adm-zip, collide with nothing, and still match on sight.

`AMBIGUOUS_EXTRACTORS` is tagged `@protocol-constant`: it is a fact about English, not a tunable vocabulary, and `archiveFunctions` is already the option for adding a real extractor.

## Test plan

- [x] `vitest` node-security: 2,823 pass, 100% coverage held.
- [x] Both real-world shapes pinned as valid cases; the existing `unsafeLib.extract` fixture renamed to `unsafeZipLib` so it still exercises the `safeLibraries` option.
- [x] Re-measured against the two repositories with the built plugin: **22 → 0** and **3 → 0**.
- [x] `rule-audit:gate --changed` — no regressions.


ofri-peretz:
`Plugin Taxonomy` is failing on `lint:detection-list-coverage`, and in this case the gate is pointing at something worth a second look rather than just asking for a test:

```
DEFAULT_NON_SECRET_TAILS → category
DEFAULT_NON_SECRET_TAILS → kinds
```

`DEFAULT_NON_SECRET_TAILS` is a **suppression** list — a name ending in one of these stops being treated as a secret. So these two entries cost **recall**, not precision: `secretCategory`, `tokenKinds` and anything else ending that way will no longer be compared safely.

That is the expensive direction. A name match that *reports* costs a user's trust; one that *suppresses* costs a real finding, silently, and nothing downstream will tell you it happened.

Two things would settle it:

1. A test per entry that **fails when the entry is removed** — that is what the gate is asking for, and it also documents why the suffix is safe.
2. A recall check in the other direction: confirm `no-timing-unsafe-compare` still fires on a genuinely unsafe comparison whose identifier ends in `category` or `kinds`. If it does not, the suppression is too broad and wants an exact-name match rather than a tail match.

Happy to be wrong about the risk if these tails only ever appear on non-secret values in practice — but that is the claim the tests should make explicit.

PR #734, opened 2026-08-25
fix(scripts): the name-inference gate read 272 rule files and skipped 207

ofri-peretz:
## The gate had a hole the size of half the ecosystem

`ruleSources()` added exactly `src/rules/<rule>/index.ts` per top-level entry. Any rule laid out as a **flat file** was never opened.

| | files |
|---|---:|
| rule files scanned | 272 |
| rule files **never read** | **207** |

Every rule under a category directory — `rules/react/hooks-exhaustive-deps.ts`, `rules/conventions/analytics-event-naming.ts`, 205 more — was invisible to this gate for its entire existence.

**This is why `hooks-exhaustive-deps` shipped name inference.** It decided which React values were "stable" from `/^set[A-Z]/`, `/dispatch/i` and `/Ref$/`, which reported real refs *and* silently exempted reactive values ([#733](https://github.com/ofri-peretz/eslint/pull/733)). The gate did not miss the pattern — **it never read the file**.

A gate reporting 272 rules clean while 207 go unread is worse than no gate, because the green tick is read as coverage.

Now walks every `.ts` under `src/rules` at any depth: **521 scanned**.

## It also catches the regex form

`P.some((p) => p.test(name))` is a dictionary lookup on a spelling exactly as much as `WORDS.some((w) => name.includes(w))` is. Only the matcher differs, and the gate could not see the difference.

## Two things only a control could tell me, both in my own work

**The first regex detector missed the exact shape it was written for.** It used `[^)]*` between `.some(` and `.test(`, and that class stops dead at the arrow's own parameter parens in `(pattern) => …`.

**Requiring the haystack to be a name cut 9 findings to 2.** Regex lists are the normal way to analyse string *content* — SQL text, a JWT header, a postMessage origin — and flagging those is 8 false alarms for every true one. A gate that noisy gets switched off, and then it protects nothing.

Three attempts at the control failed silently before one worked, because my probe file sat outside the scan — **the same blind spot as the bug**. That is the third time this session a green result meant "not measured" rather than "fine".

## The 9 newly-visible sites, registered as debt

Several are the **JSX convention**, not this defect class: `/^[A-Z]/` to tell a component from a host element is the language's own rule, and `elementName.includes('-')` is the HTML spec's definition of a custom element. They are registered because the detector cannot distinguish a convention from a vocabulary, and special-casing the gate around them would be the "write the exception into the gate's doc comment" mistake `CLAUDE.md` names.

`ddd-anemic-domain-model` is the opposite and the worst in the set: **eight `.includes()` calls** against `repository|service|api|client|dao` to decide what a class *is*. That one is the first to fix.

## Test plan

- [x] Control: the exact `hooks-exhaustive-deps` shape, planted in a **flat** rule file the gate previously never opened, is now caught.
- [x] Stale-registry direction still fails — verified by pointing an entry at a nonexistent file.
- [x] `lint:name-inference` green at 521 scanned, 48 registered.
- [x] Script suite: 1638 passed.

## After merge

Nothing ships; repo tooling. The `hooks-exhaustive-deps` entry should be deleted when #733 lands — the gate will demand it, since a stale entry fails.

## Summary by CodeRabbit

- **New Features**
  - Expanded detection of name-inference patterns, including regular-expression checks and additional vocabulary patterns.
  - Added coverage for newly identified patterns across convention, security, React, and other rule categories.
  - Broadened rule discovery to cover all applicable TypeScript rule definitions.

- **Bug Fixes**
  - Improved detection consistency across supported matching approaches.
  - Updated rule tracking to reflect newly identified detection and suppression cases.


ofri-peretz:
Both fixed, plus a third the gate surfaced once they were in.

**Two-regex threshold.** `REGEX_LIST` ended in `{1,}` over "regex-then-comma", so a single literal followed by anything qualified — your `const P = [/^x$/, dynamicPattern]` built a vocabulary from one regex, against the threshold the header documents. It now requires one-or-more `regex,` followed by a **trailing literal**, which is two minimum. Checked across four shapes: one-plus-dynamic rejects, single rejects, two accepts, three accepts.

**Identifier escaping.** Applied at all three `.join('|')` sites, including `boundSubstring`. Demonstrated rather than assumed — building the matcher both ways against `patterns$.some((p) => p.test(node.name))`:

```
unescaped matches: false   <- $ acts as an anchor
escaped   matches: true
```

This was the more dangerous of the two: it fails *silently*. The gate reports clean while the blind spot it exists to close stays open.

**Third, found by the fixes.** With those in, the gate flagged `hooks-exhaustive-deps` as a debt entry with no sites left. Its own `reason` said to delete it when #733 landed — #733 merged earlier, so the entry was stale and the bidirectional ratchet was right to fail on it. Removed.

Gate now: 521 rules scanned, no new inference, 47 known sites registered.

ofri-peretz:
Fixed in `bb4d8f61f` — `boundRegex` now uses the same escaped-character atom `REGEX_LIST` does.

One correction worth recording, because it changes what the regression test should be.

**Your example does not reproduce it.** `/foo\/bar/.test(n)` matches under the old class too — the engine backtracks and finds `/bar/` as a valid literal inside it. I wrote the regression test from that example first, and it was vacuous: it passed with the fix reverted.

```
"/foo\/bar/.test(n)"   old: true   new: true    <- proves nothing
"/\//.test(n)"         old: false  new: true    <- the real case
"/a\/b\/c/.test(n)"    old: true   new: true
```

The distinguishing case is a literal whose **only** content is the escaped slash. `/\//` has no shorter valid slice to fall back on, so the old class matched nowhere and the site went undetected. That is the test now, and reverting the atom fails it.

So the bug is real and worth fixing, but narrower than "any escaped slash" — it needs a literal with no alternative parse. Gate still green: 521 rules scanned, 46 registered.

PR #737, opened 2026-08-25
fix(modularity): ddd-anemic-domain-model matches DTO names by suffix

ofri-peretz:
## The defect

`DTO`, `Dto`, `Data`, `Request`, `Response`, `Payload` describe a **positional** convention — `OrderDto`, `CreateUserRequest`, `LoginPayload`. Matching them with `.includes()` turned each into a substring.

This is a **suppression**, so every collision costs a real finding rather than adding noise:

| class | before | after |
|---|---|---|
| `Person` | reported | reported |
| `Requestor` | **silent** — contains `Request` | reported |
| `Responder` | silent | reported |
| `DataMapper` | silent | reported |
| `OrderDto` | exempt | exempt |
| `CreateUserRequest` | exempt | exempt |

A requestor is an actor, not a data carrier. Its anemia is exactly what this rule exists to find, and it disappeared on the spelling. `Updater` survived only because it contains `Update` and not `Data` — luck, not design.

## Two things the positive control taught me

**My first five probes were all quiet, and I nearly read that as "no collisions".** They were quiet for an unrelated reason: this rule reports nothing outside a domain directory. The rule's own test fixture was quiet through the Linter too. It fires only under `domain/` or `entities/`.

That is also why it has **zero findings on the pinned corpus** — none of the eight repositories uses that layout. So this defect was invisible to every corpus-based measurement I have been running, and would have stayed invisible.

**It was invisible to `lint:name-inference` as well.** That gate only read `<rule>/index.ts`, and this rule is a flat file — one of the 207 the gate never opened. [#734](https://github.com/ofri-peretz/eslint/pull/734) widens the scan, and this is the first thing it surfaced.

## Test plan

- [x] Positive control first, and the test file says why: everything else is meaningless unless an anemic class in `domain/` actually reports.
- [x] Six genuine DTO suffixes stay exempt.
- [x] Four collisions now report: `Requestor`, `Responder`, `DataMapper`, `PayloadValidator`.
- [x] Verified failing: reverting to `.includes()` reds exactly the three collision tests.
- [x] `eslint-plugin-modularity` full suite: 175 passed, coverage at threshold.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of DTO class names by matching configured markers at the name’s suffix.
  * Anemic domain classes with names merely containing terms such as `Request`, `Response`, `Data`, or `Payload` are now correctly reported.
  * Genuine DTO names, such as `CreateUserRequest`, continue to be exempt.

* **Tests**
  * Added coverage for valid DTO suffixes and false-positive scenarios.

* **Documentation**
  * Added a changeset describing the updated behavior.


ofri-peretz:
You're right that `isDto()` decides from the class name alone, and I'm not fixing it in this PR. Here is the reasoning, because the "use explicit DTO evidence" option is narrower than it looks.

**A DTO and an anemic domain model are structurally identical.** Both are data carriers with no business methods. There is no AST-structural predicate that separates them — the difference is intent, and intent only exists in the name. So "explicit DTO evidence" reduces to decorators (present only in class-validator/class-transformer codebases) or file location, both of which silently change behaviour for everyone else.

The three real options, and their cost to consumers:

| | Effect |
|---|---|
| Drop the exemption | Every DTO under `domain/` reports. Defensible — a DTO in `domain/` is itself a smell — but noisy on day one |
| Require decorator evidence | Works for some codebases, silently changes the contract for the rest |
| Ship the suffix fix, defer the redesign | Lands a real improvement now; the contract change gets its own PR and release note |

Going with the third. This PR already fixes a live bug — `.includes()` → `.endsWith()` stopped `Requestor` being silently exempted because it contains `Request`, which cost a genuine finding on a positive control. That is worth landing on its own. Quietly redesigning a published rule's semantics inside a PR titled "matches DTO names by suffix" is not.

**Something your comment surfaced that is worth more than the fix.** `lint-name-inference` defines `SUBSTRING_METHODS` as `includes|indexOf|search|match`. It does not include `endsWith` or `startsWith` — so this PR's own fix moved the code *out of the gate's range*. The name inference is still there; the gate simply stopped seeing it.

Widening that list flags **4 further rules** across jwt-security, nestjs-security, node-security and secure-coding. That is a separate PR — registering four unrelated sites here would hold this one hostage — and I'm opening it next.

PR #739, opened 2026-08-26
fix(workspace): the precision benchmark never ran — repair it, then measure honestly

ofri-peretz:
## Summary

`benchmarks/score.ts` has never been able to invoke ESLint, and its error handling turned that outage into a published number. This repairs the harness, corrects how findings are attributed, sets a precision goal, and pins the first four real-world false positives into the corpus.

## The harness was broken, and hid it

Three defects stacked, and the third concealed the first two:

1. **Wrong cwd.** `lintFile` resolved the repo root as `'../..'` from `benchmarks/` — the directory *above* the checkout. No `node_modules/.bin/eslint` exists there, in this repo or in any worktree.
2. **Stale plugin paths.** `eslint.benchmark.config.mjs` imported `packages/eslint-plugin-pg` and `packages/eslint-plugin-jwt`; those directories are now `eslint-plugin-postgresql-security` and `eslint-plugin-jwt-security`. The config threw on load.
3. **Failure scored as a clean file.** `lintFile` ended in `catch { return 0 }`, with `2>/dev/null` on the invocation. A crashed run reported zero findings per file, so every vulnerable sample became a false negative and every safe sample a true negative:

```
Aggregate: TP=0 FN=69 TN=60 FP=0
Precision: 0%  Recall: 0%  F1: 0%
```

That is what it printed for me before this PR. **A harness that cannot distinguish "no findings" from "never ran" manufactures a number either way** — and it would just as happily have printed a perfect score. `lintFile` now throws with the captured stderr; ESLint exiting 1 because a file genuinely *has* findings is still handled normally.

## Attribution was a category error

The scorer counted every finding on a safe file as a false positive for the CWE under test. A fixture in `CWE-073/safe` asserts one thing — that it contains no CWE-073 — and says nothing about authentication or CSRF. That has a bad property: **adding correct rules degraded the score**, so the metric punished coverage.

Findings are now attributed by the CWE the rule itself stamps. Strict string equality proved too brittle, because the corpus and the rules both classify correctly and still disagree:

```
corpus CWE-1333  <-  no-redos-vulnerable-regex  stamps CWE-400  (parent)
corpus CWE-116   <-  no-improper-sanitization   stamps CWE-79   (consequence)
corpus CWE-327   <-  jwt/no-algorithm-none      stamps CWE-347  (sibling)
```

All three are **detected**. Scoring them as misses measured our taxonomy, not our rules. A finding therefore counts when its CWE matches *or* it came from a plugin the manifest already nominated in `expectedPlugins`. A stricter `fpAnyRule` count and the specific off-category findings are recorded alongside, so the looser attribution can never hide a suite that is noisy on clean code.

| scoring | TP | FN | TN | FP | P | R | F1 |
|---|---:|---:|---:|---:|---:|---:|---:|
| strict CWE equality | 59 | 10 | 60 | 0 | 100% | 86% | 92% |
| with `expectedPlugins` | 69 | 0 | 59 | 1 | 99% | 100% | 99% |
| **+ real-world fixtures (this PR)** | **69** | **0** | **61** | **3** | **96%** | **100%** | **98%** |

## Four real-world false positives, pinned

Every FP that has cost us a conversation came from third-party code, and none of those shapes were in the corpus — which is why each had to be found by hand. Each fixture carries `@source` (the repo and file it came from) and `@sealed` (the rule it holds down):

| fixture | shape | status |
|---|---|---|
| `CWE-022/safe/otel-propagation-extract.js` | `propagation.extract(context.active(), headers)` | passes — holds down #727 |
| `CWE-022/safe/oauth-claims-extractor.js` | `claimsExtractor.extract(token, 'access')` | passes |
| `CWE-079/safe/dompurify-optional-chaining.js` | `window.DOMPurify?.sanitize(html, CFG) ?? html` | **fails** — reproduces the blind spot |
| `CWE-798/safe/typeorm-migration-name.js` | `name = 'AddRewardTokenToJobTable1733127731356'` | **fails** — ~20 findings from one generated pattern |

The two failures are deliberate. `dompurify-optional-chaining` reproduces what would have shipped **21 false criticals across seven Adobe repositories**; `typeorm-migration-name` reproduces a single pattern that produced roughly twenty findings in one repo. A fixture that fails on the unfixed rule and passes on the fixed one is the standard this repo already applies to bugs.

Precision moves 99% → 96% **because the corpus got more honest.** That is the intended direction.

## The goal

`benchmarks/PRECISION_GOAL.md` sets it, and deliberately does not make the synthetic number the target:

- **Primary — real-world precision ≥ 98%**, measured on pinned third-party code with a per-rule FP budget and a CI gate, mirroring the latency ratchet that already works (`check-per-rule-budget.ts` + `per-rule-budget.yml`).
- **Secondary — zero net-new suppressions.** `eslint-disable` comments naming our rules in adopter repos is precision judged by the only party that counts. IGNF wrote **36**, each with a hand-written justification. A synthetic corpus can be gamed by editing fixtures; a suppression count cannot.
- **Coverage** — 31 CWEs have fixtures against **374 rules**. A rule with no fixture has no precision data at all, and the scorecard already flags these `⚠️ none`. M4 closes that column.

## Verified

- `npx tsx benchmarks/score.ts` — runs end to end for the first time; every number above is reproducible from this commit.
- The one pre-existing corpus FP is left in place: `CWE-327/safe/verify-allowlist.js` trips `jwt/require-issuer-validation`, `require-audience-validation` and `require-max-age`. The fixture only ever claimed a safe algorithm allowlist, so this is our JWT rules asking for more hardening than it set out to show. That is a decision to make, not a fixture to quietly edit.
- `benchmark-results/history.ndjson` is deliberately untouched — my exploratory runs included two fabricated 0% entries and do not belong in the record.
- Two pre-push hooks excluded for worktree-local reasons (`recheck` absent; Turbopack rejects the symlinked `node_modules`); this PR touches no rule source.

## Summary by CodeRabbit

* **Documentation**
  * Added benchmark goals covering precision, recall, suppression rates, evidence coverage, milestones, fixture provenance, and known issues.

* **Benchmark Improvements**
  * Added safe security fixtures covering OAuth claims, OpenTelemetry propagation, DOM sanitization, and database migrations.
  * Added comprehensive CWE benchmark results and improved scoring accuracy, error reporting, and finding classification.
  * Updated security plugin configuration to use current package paths.


ofri-peretz:
All three fixed in `3267aaab1`, plus the conflict with `main` resolved.

**Fatal messages scored as false negatives.** This was the one that mattered most for a PR titled "measure honestly". A fatal message arrives with `ruleId: null`, which mapped to `{ ruleId: null, cwe: null }` — indistinguishable from *"the rule looked and found nothing"*. So a vulnerable fixture that failed to parse was counted as a **miss**, and a syntax error anywhere in the corpus silently lowered measured recall while reading as a rule defect. It is the same failure class as the `catch { return 0 }` this file already documents in its own comment: an outage wearing the costume of a result. It now throws and names the fixture and the parse error.

**Shell interpolation.** `execFileSync` with an argument array, so the filename is an argument rather than a fragment of a command line. `ESLINT_USE_FLAT_CONFIG` moves to the child's `env` for the same reason — with no shell there is nothing to interpret `VAR=x cmd`. Worth noting *why* this one is not theoretical: a benchmark corpus is precisely where a hostile filename would arrive, because adding a fixture looks like contributing data rather than contributing code.

**Unpinned `@source`.** All five repositories now pinned to a commit resolved from the GitHub API:

```
aemdemos/lundbeck-vyepti        @323f9ce8…
auth0/express-openid-connect    @9cdf9844…
diia-open-source/be-diia-app    @51b50099…
humanprotocol/human-protocol    @114980f1…
vitemcp/server                  @d438700e…
```

A citation without a SHA cannot support the claim it exists for — the referenced code can change, so a reviewer checking a claimed false positive may find something different from what was measured. Since these citations back a *published precision number*, that is the first thing a sceptical reader checks.

`benchmarks/__tests__/fixture-source-pins.test.ts` locks it. Verified by sabotage — removing one SHA fails it — and it also asserts it found citations at all, so pointing it at the wrong directory cannot pass it silently.

One thing I could not verify locally: the full scorer needs a built devkit, and this worktree's `dist/` is stale (`createModuleEvidence is not a function`). The export chain is intact on `main`, so CI's clean build is the real check. Notably the harness threw loudly rather than scoring zero — which is the behaviour this PR added.

ofri-peretz:
Pushed `85fc2eb53` — this is what was failing `Script & Repo-Config Locks`.

`check-intent.ts` and `check-new-rule-cases.ts` each export helpers *and* run their gate at module top level, and the gate ends in `process.exit(1)`. So `check-intent.test.ts` importing `parseFrontmatter` executed the whole gate and took the vitest worker down with it:

```
Error: process.exit unexpectedly called with "1"
```

That failed every test in the file, including the ones that never touched the gate. The tell was that both files **passed when run alone and failed in the full suite** — the signature of an import-time side effect rather than a broken assertion, which is why chasing the assertions led nowhere.

Each executable body is now wrapped in an `IS_ENTRYPOINT` check — the same shape `lint-changesets.ts` already uses. Exports untouched; both CLIs still run and exit 0 with identical output.

`scripts/__tests__/script-entrypoint-guards.test.ts` locks it, and asserts two things rather than one: that the guard exists, and that it **names its own filename**. A guard checking the wrong file never fires but still looks guarded to any scan for the shape.

Verified by sabotage:

```
guard removed ............... 2 assertions fail
guard names the wrong file .. 1 assertion fails
restored .................... 4 pass
```

Worth flagging: my first sabotage attempt *passed*, and I nearly took that as proof. Prettier had reflowed the constant across three lines, so the string I was replacing never matched and I had tested nothing. The check is only worth something if the broken version actually fails.

Two `check-intent` tests also fail intermittently on my machine with `Hook timed out in 10000ms` inside `beforeEach` — that is temp-git-repo setup under load, not the guard; a re-run passed 14/14. If CI shows it too, the hook timeout wants raising, but that is a separate change.

PR #695, opened 2026-08-24
chore(release): version packages

ofri-peretz:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and publish to npm yourself or [setup this action to publish automatically](https://github.com/changesets/action#with-publishing). If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases
### Major Changes

- **💥 Breaking** — `recommended` no longer enables `no-magic-numbers` ([#694](https://github.com/ofri-peretz/eslint/pull/694))

  It produced **1421** findings on the pinned 8-repository corpus — the largest
  single source in the ecosystem — and the project's own triage records it as
  "correct in contract, and a taste rule by nature". Those two facts together are
  the argument: a taste rule on by default is what makes a consumer stop reading
  the output, and an ignored tool has zero recall regardless of what it detects.

  This is parity with upstream, not a novel opinion. ESLint core ships
  `no-magic-numbers` and deliberately excludes it from `eslint:recommended`.

  **Nothing is removed from the plugin.** The rule is still exported, documented
  and tested — only its membership in `recommended` changed.

  ## Migration

  If you were relying on `recommended` to enable it, add one line. Most projects
  want the opposite and need to do nothing.

  ```js
  // Before — enabled implicitly by the preset
  export default [conventions.configs.recommended];

  // After — enable it explicitly
  export default [
    conventions.configs.recommended,
    { rules: { 'conventions/no-magic-numbers': 'warn' } },
  ];
  ```

  Same reasoning as `eslint-plugin-import-next` 2.4.0, which dropped `order`,
  `first` and `newline-after-import` from its `recommended`.

### Minor Changes

- **✨ Feature** — Three rules no longer report inside generated files: `conventions/no-magic-numbers`, `conventions/no-commented-code` and `modernization/prefer-template-literal`. ([#691](https://github.com/ofri-peretz/eslint/pull/691))

  Each one's remedy is unavailable in a file a generator owns. "Name this
  constant", "delete this commented-out code" and "use a template literal" all
  mean _edit this file_ — and the next codegen run overwrites the edit. The
  finding is correct and unactionable, which is the definition of an effective
  false positive.

  `no-magic-numbers` alone is the largest single source of findings on the pinned
  corpus at 1421, and the project's triage already recorded it as "correct in
  contract, and a taste rule by nature".

  This is not a blanket exemption for generated code. A rule whose finding
[...]

PR #748, opened 2026-08-27
feat(release): CS009 catches a changeset that would mis-scope its release notes

ofri-peretz:
## The defect this prevents

Changesets applies a changeset's body **verbatim to every package it lists**. A changeset covering `eslint-plugin-jwt-security` and `eslint-plugin-node-security` whose body describes rules from both writes both paragraphs into both CHANGELOGs — so each package's published release notes claim rules it does not own.

**That shipped.** It reached [#695](https://github.com/ofri-peretz/eslint/pull/695) and was caught in review, not by tooling. Both existing gates passed it:

- `lint-changesets` (CS001–CS008) — checks prose quality, not scope
- `verify-release-notes` (RN001–RN006) — checks structure, install lines, code spans

Neither can express *"this entry names another package's rules"*. And RN runs **after publish**, so even a check there would have been describing something already on npm.

## How CS009 decides

Ownership is read from the `packages/` tree, not a hand-kept list — a rule belongs to whichever `eslint-plugin-<x>/src/rules/<name>/` declares it. The **14 rule names carried by more than one plugin are dropped**: `no-hardcoded-secrets` exists in several, so seeing it says nothing about which package a sentence describes, and guessing would manufacture false positives.

It reports only when a multi-package changeset names rules owned by **two or more of the packages it lists** — the precise condition under which the shared body becomes wrong for someone.

Verified against the real thing:

```
❌ CS009  regression.md
   Covers 2 packages and describes rules from 2 of them (jwt-security,
   node-security). Changesets copies this body into every listed package's
   CHANGELOG, so each would advertise the others' rules. Split into one
   changeset per package.
```

## Deliberately quiet

Each of these has a test, because a scope check that cries wolf gets disabled:

- multi-package changeset naming **no** rules (shared infra change)
- multi-package naming rules from **only one** listed package
- **any** single-package changeset — one package cannot contaminate another, so a foreign rule name there is a cross-reference
- rule names appearing as **prose** rather than in backticks

## Test plan

- [x] `vitest run scripts/__tests__/lint-changesets.test.ts` — **36 pass**.
- [x] **Verified non-vacuous**: disabling the check (`if (false)`) fails exactly one case; restoring it passes all 36.
- [x] Ran against the real `.changeset/` — clean, no false positives on live changesets.
- [x] Typecheck, prettier, markdownlint clean.

## Note on the test fixtures

The CS009 cases use **real** plugin and rule names, unlike the rest of the file. Ownership is the thing under test and it is read from the actual tree — a synthetic `eslint-plugin-published` owns nothing, so a fixture in the house style could never fire and would pass however broken the check was.

## Summary by CodeRabbit

* **New Features**
  * Added validation for multi-package release notes that reference rules from multiple packages.
  * Reports an error when backticked rule references span unrelated package ownership.
  * Correctly allows shared rules, single-package changesets, and valid one-package references.

* **Documentation**
  * Added documentation for the new cross-package scope validation rule.


ofri-peretz:
Fixed in `4dd0c5e68`. Good catch, and the failure mode is the bad one.

`indexOf('/')` on `@interlace/eslint-plugin-jwt-security/no-decode-without-verify` left `eslint-plugin-jwt-security/no-decode-without-verify`, which matches no bare rule key. So the owner lookup returned nothing, `named` stayed empty, and CS009 **passed a genuinely cross-scoped changeset in silence** — no error, no warning, nothing to notice. For a check whose entire purpose is catching what review already let through once, a false negative just restores the hole it was written to close.

Now `lastIndexOf('/')`, which handles all three forms uniformly: bare `no-decode-without-verify`, plugin-qualified `jwt-security/no-decode-without-verify`, and scoped `@interlace/eslint-plugin-jwt-security/no-decode-without-verify`.

Added the case you asked for — two scoped references across two listed packages. Reverting to `indexOf` fails it (and one other), so it is not vacuous. 37 tests pass.

PR #749, opened 2026-08-30
feat(workspace): map every plugin name and link, add theme-aware brand marks

ofri-peretz:
Follow-on to #747. Same theme: a plugin has **eight identifiers stored in eight files**, and nothing has ever put them side by side.

## The map

`.agent/link-and-name-map.md`, generated by `scripts/map-links-and-names.ts` from the sources of truth — the docs registry, `packages/`, the prefix-identity lock's alias table, `ECOSYSTEM_LOGO`, `codecov.yml`, and the images and logos on disk.

Per plugin: workspace directory · npm name · rule-id prefix · deprecated alias · docs slug + pillar · OG banner · ecosystem logo · codecov component. Then every URL shape we author, and which gate holds each column.

**Derived, never hand-maintained.** Every hand-kept plugin list in this repo has drifted at least once — the `DESCRIPTIONS` map, `VALID_PLUGINS`, the README ecosystem table, the codecov flags. A map that needs updating is one more of those. When it disagrees with a source file, the source file is what changes.

### It earns its place immediately

**Eleven plugins have no codecov component** — `anthropic`, `drizzle`, `gemini`, `knex`, `mcp-sdk`, `mysql`, `openai`, `prisma`, `sequelize`, `sqlite`, `typeorm`. Each of their READMEs already renders a Codecov badge pointing at `component=<pkg>`, and I confirmed against codecov that those resolve to **`unknown`**.

Reported, not fixed: `codecov.yml` targets 100% project + patch per component, so whether these eleven are ready for that gate is a coverage decision, not a naming one.

## Theme-aware brand marks

Interlace, ESLint and oxlint appear in every logo row, and one file cannot read on both npm (always white) and GitHub dark. Each now ships a `-light` / `-dark` pair from `tools/scripts/make-theme-variants.mjs`, written into every README as `<picture>` with the `-light` file as the `` fallback — so a renderer that drops `<source>` still gets the variant that suits npm's white background.

Nothing invents a colour: Interlace already shipped both pairs as first-party artwork, ESLint's dark treatment is copied out of the `prefers-color-scheme` query in its own file, and oxlint's two are the vendor's own files relabelled.

**Ours are named by SURFACE, not by ink** — `-light` means *for light surfaces*. That is the opposite of oxc's convention, and the collision has already caused one wrong-variant commit here, so the generator, the rule doc and the map all repeat it.

Verified by rendering all six on white and on `#0d1117` and looking at them, per the `_verify` note in `logo-sources.json`. The wrong-pairing row makes the case: oxlint's wordmark disappears entirely.

## Shorter still

The why/how/what from #747 is now **three bullets and a link line, under no heading of its own**. Three `##` sections for one thought was three table-of-contents entries between a reader arriving from npm and the install command. The reference is the NestJS npm README.

## Test plan

- [x] `check:readme-structure` — all 31 READMEs canonical.
- [x] `logos:variants:check` — 6 variants current.
- [x] `map:names:check` — map in sync.
- [x] `lint:md` 0 issues; `sync:root-readme:check` in sync; `sync-readme-rules.ts` idempotent (0 of 30 modified on re-run).
- [x] **New lock** `scripts/__tests__/link-and-name-map.lock.test.ts` runs the generator's `--check` from vitest, so a rename moves the map in the PR that makes it. Verified non-vacuous — editing one cell turns it red.
- [x] Structure gate extended: requires the `-light`/`-dark` pair, rejects the theme-agnostic base file in a README.

All three new checks are also chained into `npm run quality`.

## After merge

The eleven missing codecov components are the open decision. Nothing else here changes runtime behaviour — it is metadata, assets and gates.


ofri-peretz:
Fixed in `425dba096`.

`existsSync(OUT) ? read(OUT) : null` acts on the result of a check, and the file can be replaced or removed in the window between the two calls — CodeQL's `js/file-system-race`.

Replaced with an attempted read that handles `ENOENT`. There is no window: the syscall either returns the contents or tells us the file is gone.

Worth noting the second bug the `existsSync` form was hiding. It treated *every* failure as "file absent" — a permissions error, or a directory sitting at that path — and the script would then have reported "missing" and overwritten it. The new form re-throws anything that is not `ENOENT`, so those surface instead of being papered over.

Both paths exercised: with the file present `--check` reports in sync; with it deleted `--check` reports missing and exits non-zero; regenerating produces a byte-identical file.

PR #751, opened 2026-08-30
fix(ci): unblock three silent health crons, define how docs quality is measured

ofri-peretz:
Answers "how do we ensure our links, naming and data stay up to date" by first measuring it — and the measurement found three refreshers that had been dead for months.

## What was broken

`commitlint.config.mjs` derived valid scopes by scanning `packages/`, `apps/` and `tools/` **literally**. `benchmarks` is a workspace member too — it is in `package.json#workspaces` and publishes as `@interlace/benchmarks` — but it is a bare directory rather than a glob, so it was never scanned and `bench` was not a valid scope.

Three workflows commit `chore(bench): …`, hit the rule, and exit 1:

| Workflow | Failing since | Consequence |
| :--- | :--- | :--- |
| `peer-health.yml` | ≥ 2026-07-20, **every weekly run** | `benchmark-results/peer-health.{json,md}` — which backs `distribution/ECOSYSTEM_LANDSCAPE.md` — frozen for six weeks |
| `resource-profile.yml` | ≥ 2026-06-01, **every monthly run** | peak RSS + cold-start numbers frozen for three months |
| `weekly-benchmark.yml` | same scope in its PR title | — |

Nobody noticed because **a red cron notifies nobody by default**. The work succeeded every time; only the commit was rejected, so the snapshot silently stopped moving. A snapshot that stops moving is indistinguishable from one that has not changed.

Scopes now come from `package.json#workspaces`, so a new workspace member is a valid scope the day it is added rather than the day someone remembers this file. The three workflows use `benchmarks`; no alias invented.

## The feedback loop was wired backwards

`integration-health.yml` already had an issue-filing step — guarded by `inputs.create-issue != false`. `inputs.*` is empty on a `schedule` run and GitHub coerces `''` to false, so the guard evaluated **FALSE on every cron**. The issue was filed only on a manual dispatch: only when someone was already watching.

It went red on 2026-08-24 and again on 2026-08-29 and opened nothing.

### ⚠️ That probe is right, and it found a live production defect

> `NEXT_PUBLIC_POSTHOG_KEY` is marked **Sensitive** in Vercel for `serverless.interlace.tools`, so the build inlined the literal string `[SENSITIVE]` instead of the key. The same var feeds `posthog-js`, so **that site is sending no analytics at all**, and its CSP report-uri is dead.

Fixing it needs Vercel access (different repo, not something I can do here): remove the var and re-add it with `--no-sensitive`.

`peer-health`, `resource-profile` and `check-links` now file an issue on failure too — one issue, reopened and commented, the shape `integration-health` already used.

## docs/DOCS_QUALITY.md

What "good documentation" means here, as **rows with a metric, a target, the check that holds it, and a cadence** — never an intention. Four sections: links, names, data, content.

It states the tiering rule that was implicit:

| The check… | goes in | because |
| :--- | :--- | :--- |
| reads only repo files | PR gate (vitest) | deterministic; the PR that breaks it is the cheapest place to fix it |
| touches the network | cron that files an issue | an upstream 404 is not the PR author's fault and must never wedge a merge |
| needs a build or benchmark | `ready_for_review` | too slow per-push, too important to skip before merge |

And one thing worth knowing on its own: **`npm run quality` is a local convenience command that no workflow invokes.** A check added only there runs when someone remembers to type it. Checks belong in `scripts/__tests__/` (run by `test:scripts` in `quality.yml`) or in a workflow step.

### Measured, not asserted — 8 gaps today

- 4 checks that exist but run only when typed: `check-links`, `check:utm-links`, `audit:claims`, `check:audit-freshness`
- **19 of 30** plugins have a codecov component (the other 11 render a badge that resolves to `unknown` — same finding as #749)
- **6 of 17** scheduled workflows have no failure channel at all: `codecov`, `cve-latency`, `eslint-version-matrix`, `oxlint-parity`, `quality-full`, `weekly-corpus-scan`
- the two revived refreshers have to prove a green run before their column can be trusted

The number to move is **gaps → 0**.

## Test plan

- [x] `npm run lint:workflows` — 35 workflow files pass conventions.
- [x] `npm run lint:md` — 0 issues in 1095 files.
- [x] `vitest run scripts/__tests__` — 1654 passed / 61 files.
- [x] Commitlint config exercised directly: `benchmarks` now resolves, `bench` correctly does not, and the 74-scope list still contains `workspace`, `docs`, `ui` and every package short form.
- [x] All four health workflows parsed with `yaml` to confirm the issue-filing step is present and its `if:` guard fires on `schedule`.

## After merge

Watch the next scheduled runs actually go green — Fri (`check-links`), Sat (`integration-health`), Sun (`peer-health`), 1st of month (`resource-profile`). `integration-health` will stay red and now file an issue until the Vercel env var is fixed; that is the correct behaviour, not a regression.

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -21,6 +21,7 @@ concurrency:
 permissions:
   contents: read
   pull-requests: write
+  issues: write
 
 jobs:
   gate:
@@ -118,3 +119,25 @@ jobs:
         run: |
           echo "❌ Link check failed. See report for details."
           exit 1
+
+      # On a PR the failure is visible: the job goes red and the step above comments
+      # the report. On the weekly cron nobody is looking, so a broken link would sit
+      # in a red run until someone happened to open the Actions tab. One issue,
+      # reopened and commented rather than a new one per week.
+      - name: 🚨 Open an issue on a scheduled failure
+        if: failure() && github.event_name == 'schedule'
+        env:
+          GH_TOKEN: ${{ github.token }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+        run: |
+          set -euo pipefail
+          TITLE="Link integrity check failed"
+          EXISTING=$(gh issue list --state open --search "$TITLE in:title" \
+            --json number --jq '.[0].number // empty')
+          BODY=$(printf '%s\n\n[Run](%s) — the full report is the `link-report` artifact.\n' \
+            "The weekly link check found broken or insecure links in the docs content." "$RUN_URL")
+          if [ -n "$EXISTING" ]; then
+            gh issue comment "$EXISTING" --body "$BODY"
+          else
+            gh issue create --title "$TITLE" --body "$BODY"
+          fi
--- a/.github/workflows/integration-health.yml
+++ b/.github/workflows/integration-health.yml
@@ -88,7 +88,14 @@ jobs:
           } >> "$GITHUB_STEP_SUMMARY"
 
       - name: 🚨 Open an issue on failure
-        if: steps.probe.outcome == 'failure' && inputs.create-issue != false
+        # `inputs.*` is empty on a `schedule` run, and GitHub coerces '' to false, so
+        # `inputs.create-issue != false` evaluated FALSE on every cron — the issue was
+        # filed only on a manual dispatch, i.e. only when someone was already watching.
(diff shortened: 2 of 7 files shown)


ofri-peretz:
Superseded by #754, which contained this branch and is now merged. Verified every file this PR touched is byte-identical on `main`: `commitlint.config.mjs`, `peer-health.yml`, `resource-profile.yml`, `integration-health.yml`, `check-links.yml`, plus `docs/DOCS_QUALITY.md`. Closing rather than merging to avoid a no-op commit.

PR #754, opened 2026-08-30
feat(workspace): close every open stage of the AI-native SDLC

ofri-peretz:
Closes every open stage of [AI_NATIVE_SDLC.md](https://github.com/ofri-peretz/eslint/blob/main/AGENTS.md). **Includes #751** (merged into this branch, so review that one first or merge this and close it).

Adapted from Anthropic's [AI-Native SDLC Playbook](https://academy.claude.com/courses/ai-native-sdlc-playbook). Each stage is built to be *used*, not described.

## Stage 1 + 2 — `intent.md` / `spec.md`

`intent/` with a README, both templates, and `intent-artifacts.lock.test.ts`.

The load-bearing rule: **an `approved` intent with no `spec.md` beside it fails the lock.** Approving an intent is supposed to mean it has been designed, not that somebody liked it — and the failure was silent, because a missing file looks like nothing at all.

## Stage 4 — continuous evals

`evals/`, two layers.

**Layer 1 — deterministic, free, always runs.** Every relative link in each of 63 agent-facing documents must resolve. It was red on its first run:

- `.agent/flagship-rules.md` pointed at an article that does not exist in this repo
- six philosophy links (`LAYOUT`, `LOADING`, `MOTION`, `PAGINATION`, `URL`, `CODE_EXAMPLE`) were never repointed after the interlace repo split

A rule document that points at a moved file is a rule the agent silently cannot read.

**Layer 2 — task evals.** Runs cases against the live configuration; needs `ANTHROPIC_API_KEY`. Without one it reports **skipped, never failed** — a fork must not be blocked by an eval it cannot run. Seeded with three cases from real incidents here, including the rename that left dead references in eight config files.

`evals.yml` fires on any change to `CLAUDE.md`, `AGENTS.md`, `.agent/**` or `lefthook.yml`, plus weekly.

## Stage 5 — the *ask* gate

Every hook here could allow or block. None could **ask**, which is what release gating actually needs.

`.claude/hooks/release-gate.sh` pauses a production deploy, a direct `vercel --prod`, and an `npm publish` — quoting the command back so a human approves a specific thing rather than a category. `RELEASE_APPROVAL=1` pre-approves a session.

The lock pins the quiet cases as hard as the loud ones, and earned it immediately: it caught the gate firing on `npm publish --dry-run`. A gate that cries wolf gets approved without reading, which is worse than no gate.

## Stage 6 — bands, and the write-back

`scripts/control-bands.ts`: mean and population σ over a rolling window, all four Western Electric rules, 1σ/2σ/3σ tiering. **No model runs in the detection path** — the decision that something is wrong has to be reproducible before it can invoke an agent unattended.

| Breach | Response |
| :--- | :--- |
| 1σ | log only (never gates) |
| 2σ | diagnose read-only |
| 3σ | may act, via PR or runbook |

Rule 4 is the point: **eight consecutive points on one side of the mean is drift no threshold can see.** Every watcher this repo had was pass/fail.

On a 2σ+ breach, `control-bands.yml` writes the diagnosis as `intent/<slug>/intent.md` and **opens a PR, not an issue** — an issue is the separate ticket queue the loop exists to avoid. Verified end-to-end against a synthetic breach: the generated intent passes the Stage 1 lock.

## Also closes the DOCS_QUALITY gaps

- **`check-links`, `check:utm-links`, `audit:claims`** now run on the PR gate in a new `docs-integrity` job. `check-links` was red — **14 dead relative links**, all fixed (six were the philosophy docs above; five were a wrong `../../` depth on `ANALYSIS-LIMITS.md`).
- **`check:audit-freshness`** runs weekly in the maintain lane. It was correct all along: it had been flagging the stale peer-health snapshot the entire time nobody ran it.
- **The six scheduled workflows with no failure channel** now share `.github/actions/report-failure` — one tracking issue per title, commented on thereafter rather than refiled weekly.

**Scorecard: 8 gaps → 1.** The remainder is the eleven plugins without a codecov component, owned by `chore/coverage-100` — components target 100%, so adding one for a package below that turns Codecov red instead of measuring anything.

## Test plan

- [x] `vitest run scripts/__tests__` — **1676 passed / 64 files** (22 of them new).
- [x] `lint:workflows` — 37 files pass. Every touched workflow re-parsed with `yaml` to confirm the step landed in the right job with the right permissions.
- [x] `lint:md` — 0 issues in 1098 files.
- [x] `check-links`, `check:utm-links`, `audit:claims`, `evals:config` — all green.
- [x] Control bands: statistics unit-tested (population σ, the σ=0 flat-series case, each Western Electric rule, and the quiet cases); `--record` verified idempotent and back-filling; write-back verified against a synthetic 3σ breach and the output passes the intent lock; `--check` exits 1 at 2σ+ and 0 at 1σ.
- [x] Ask-gate: 10 cases, both directions, including `git push`, `npm run build` and `npm publish --dry-run` passing straight through.

## After merge

Nothing here changes runtime behaviour — it is artifacts, gates and watchers. What it does not yet have is **history and usage**:

- every band holds 3 observations and computes nothing until 8; that is the honest state, not a bug
- the eval suite has 3 cases against the playbook's target of 20–50, one per incident
- no hand-written `intent.md` exists yet — the first real test of Stage 1 is the next substantial piece of work

`ANTHROPIC_API_KEY` needs to exist as a repo secret before eval layer 2 does anything.


ofri-peretz:
## Follow-up — every stage is now loaded, not merely wired

The first commit wired the stages; this one makes each strong enough to trust. A stage that is present but empty reads as coverage it does not have.

### Stage 6 — bands that actually compute

They held **3 observations each** and could say nothing until 8. The suites prune dated result files (`check:benchmark-results-size`), so the tree held three while **git held twelve**. `--backfill-git` walks the deleted ones.

| Band | Points | State |
| :--- | ---: | :--- |
| `cwe-corpus-f1` | 11 | ✅ computing |
| `cwe-corpus-recall` | 11 | ✅ computing |
| `cwe-corpus-precision` | 11 | ✅ computing |
| `oxlint-parity-rate` | 8 | ✅ computing |
| `interlace-arena-violations` | 6 | says so, computes nothing |

`actions/checkout` is pinned to `fetch-depth: 0`. On the default shallow clone `git log --all` finds nothing and the backfill returns **zero observations silently** — the bands would never reach `minPoints` and nobody would know why. That is the exact silent-zero class this repo keeps rediscovering, caught before it shipped rather than after.

### Stage 4 — twenty eval cases

Up from three, the playbook's lower bound. Each is drawn from a real incident here — the rename that left dead references in eight config files, `npm run quality` not being a CI gate, a network check on the merge path, a hand-maintained plugin list drifting.

`eval-cases.lock.test.ts` pins the corpus, and its most useful assertion is that **a prompt must not contain its own expected answer**. It caught six of my first drafts: every one framed *"answer yes or no"*, which hands the grader the word `no`. Those six now grade on the rationale, which is what the eval was supposed to measure in the first place.

### Stage 1/2 — two real intents

Both for decisions genuinely open right now, so Stage 1 has been used rather than just templated:

- **`intent/cwe-089-corpus-driver-evidence/`** — the corpus predates evidence-gated detection, with the side-by-side measurement and three open questions.
- **`intent/codecov-components-for-every-plugin/`** — the eleven `unknown` badges, and it carries its **`spec.md`**: requirements R1–R5, the design, the verification commands, and the rejected alternatives with reasons.

### Stage 3 — the contract is in the repo

`CLAUDE.md` now carries the six stages and the rules easiest to get wrong: never widen a band to clear a breach, a breach becomes an intent and not a ticket, editing `.agent/**` is a configuration change. It lived only in the parent directory before, so an agent grepping this repository could not find it.

### Verification

- 1700 tests / 65 files pass · 37 workflows lint clean · markdown clean
- `turbo run build` + `turbo run test --filter=...[origin/main]` + `oxlint:shims:verify` — exit 0
- `npm run bands` green on real recovered history

### Still needed after merge

`ANTHROPIC_API_KEY` as a repo secret, or eval layer 2 stays skipped forever and only the free layer runs.

PR #757, opened 2026-08-30
fix(ci): keep the changesets action major, its inputs and the CLI major together

ofri-peretz:
## Releases are blocked right now

The Version PR stopped refreshing. A grouped Actions update moved `changesets/action` to **v2.1.1** while the comment above it still read `v1.9.0`, and v2 renamed three inputs:

```
Error: The following inputs have been renamed:
- "version" -> "version-script"
- "commit"  -> "commit-message"
- "title"   -> "pr-title"
```

The step hard-errors, so `main` has been carrying a red check and no Version PR is being created.

## This is the second time

The first (#579) was **silent** — the Version PR simply stopped being created, which is indistinguishable from "nothing to release". Two customer-facing fixes sat unpublishable behind it before anyone noticed.

This time it errored loudly, which is luckier rather than safer. Releases were blocked either way.

## Why not just rename the inputs

Because that leaves **v2 running against `@changesets/cli ^2.31.1`**, which it does not support — v2 requires CLI v3. Renaming would turn a loud failure back into the silent one. The pin goes back to v1.9.0.

## The comment was the guard, and it failed twice

Both times the only thing standing between a grouped update and a blocked release was a comment asking the next bump to "confront the constraint rather than sail past it". Both times it sailed past.

Two mechanisms replace it:

- **`.github/dependabot.yml`** now ignores `version-update:semver-major` for `changesets/action`, so the group cannot carry it across again.
- **`scripts/__tests__/changesets-action-pin.test.ts`** asserts the SHA, that there is exactly one reference to the action, that the **v1 input names** are present and the v2 ones absent, and that the dependabot entry exists.

That last assertion matters on its own: the v2 SHA *with* v1 input names is a workflow that cannot run at all, and this catches that combination directly.

## Test plan

- [x] `vitest scripts/__tests__/changesets-action-pin.test.ts` — 4 pass.
- [x] **Verified by sabotage**: re-bumping the SHA to v2.1.1 fails two assertions; restoring passes all four.
- [x] `lint-workflows` — 35 workflows pass; both YAML files parse.
- [x] Confirmed the v1.9.0 SHA against the upstream tag (`3841a068…`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored the release workflow’s Changesets action to a compatible v1.9.0 version.
  * Prevented incompatible major updates that could disrupt version pull request creation.

* **Tests**
  * Added automated checks to ensure the action remains pinned correctly, uses supported inputs, and is protected from unintended major upgrades.


ofri-peretz:
You caught me making the wrong fix. Corrected in `5a45934aa`.

`package-lock.json` resolves `@changesets/cli` to **3.0.1**, so **v2.1.1 is the correct major** and the bump was right — only the inputs were missing. I had pinned *back* to v1.9.0, which would have paired a v1 action with a v3 CLI and kept releases broken while looking deliberate.

What led me there: the comment above the pin stated the repo was on CLI `^2.31.1` and that the pin must stay on v1.9.0. That was true when written and stopped being true when the CLI moved to v3. I also confirmed `2.31.1` from `npx changeset --version` — which read a stale local `node_modules`, not what CI installs. **The lockfile was the authority and I checked it second.**

Now:

- **v2.1.1 restored**, with `version-script` / `pr-title` / `commit-message`.
- The stale comment is replaced with what is true, plus why it misled.
- The guard is a test rather than prose, and asserts the **pairing** — action major, the input names that major accepts, and the CLI major read from the lockfile.

On your second point, the `version` assertion is now scoped to the slice after `uses: changesets/action@` rather than the whole file, and it checks the names *for the pinned major* instead of hardcoding one set.

Verified against both drift directions:

```
v1 SHA against CLI v3 ......... 2 assertions fail
v2 SHA with v1 input names .... 1 assertion fails   (this outage)
correct pairing ............... 4 pass
```

The dependabot ignore stays. It stops a grouped update carrying the action across a major on its own; upgrading is now a deliberate three-line change — SHA, input names, CLI — and the test fails until all three agree.

PR #819, opened 2026-09-02
docs(node-security): record why detect-non-literal-fs-filename's own safePattern looks like an FP

ofri-peretz:
## Summary

`detect-non-literal-fs-filename` prints `path.resolve(SAFE_DIR, path.basename(userInput))` as its `safePattern` for several fs methods, and probing exactly that shape produces a finding. That reads as a rule reporting its own advice — the unsatisfiable case its header explicitly warns about — and the obvious next move is to weaken the `path.basename` sanitiser.

**It is neither of the two readings.** The taint does not survive the `path.resolve` wrapper, and the rule is not refusing to reason about `basename`. Measured with the rule loaded directly:

| snippet | findings |
| --- | --- |
| `fs.readFileSync(path.resolve(SAFE_DIR, path.basename(req.query.f)))` — `SAFE_DIR` undeclared | 1 |
| same expression, `const SAFE_DIR = '/uploads'` above it | 0 |
| same expression, `import { SAFE_DIR } from './config'` | 0 |
| same expression, `SAFE_DIR` a function parameter | 0 |
| **`fs.readFileSync(path.resolve(SAFE_DIR, 'notes.txt'))` — undeclared, NO taint at all** | **1** |

The last row is the discriminator. There is no attacker-reachable value anywhere in it and it reports identically, so what fires is `containsFreeVariable` on a name bound nowhere in the file — the documented mechanism, working — and `readsTaintSource` never sees past `basename`.

So no behaviour changes, and `safePattern` is left alone: the advice is accepted verbatim in every file where its base actually exists, which is every real file, since an undeclared `SAFE_DIR` is a `ReferenceError` at runtime and a `no-undef` error at lint time. What this PR ships is the diagnosis, so the next person probing a fragment does not re-derive it and "fix" the sanitiser.

- Header block gains a `✗` entry with the four measurements and the discriminator.
- `path-guards.test.ts` pins both directions: two valid cases for the remediation (declared base, imported base) and the literal-second-argument CONTROL beside them.
- Changeset (patch) carries the same discriminator into the release notes.

## Test plan

- [x] `npx vitest run --coverage.enabled` in `packages/eslint-plugin-node-security` — 102 files, 2894 tests, **100%** statements / branches / functions / lines.
- [x] **Mutation-verified, both directions.** Returning `true` from the `path.basename` branch of `readsTaintSource` fails 3 valid cases including the two new ones; short-circuiting `containsFreeVariable` fails exactly the new CONTROL. Neither lock is vacuous.
- [x] `npx tsx scripts/rule-audit-gate.ts` — 121 rules, no regressions.
- [x] `npm run lint:name-inference` — 524 rules, no new name-substring inference.
- [x] `npm run check:case-registry` — registry clean, 137 verified, 0 regressed.
- [x] `npx tsx scripts/lint-changesets.ts .changeset/*.md` — passes.
- [x] `prettier --check`, package build, commit/push hooks (no `--no-verify`).

Named in the task as `npm run check:computed-keys`: that script does not exist at the root. The nearest gates ran instead — `computed-key-paths.test.ts` (inside the suite above) and `lint:name-inference`.

## Pre-existing red on `origin/main`, not introduced here

Two hooks were excluded at commit/push with the justification recorded in the commit message. Both are red on `origin/main` and cannot be moved by a change confined to `packages/eslint-plugin-node-security`:

- **`docs#test`** — `apps/docs/src/__tests__/shimmer-button-effects-lock.test.tsx` imports `@interlace/ui/effects/shimmer-button`, a subpath `packages/ui/package.json` does not export on `origin/main` either (`git show origin/main:packages/ui/package.json | grep -c shimmer-button` → 0).
- **`portability`** — hash drift on the installed oxlint runtime files (`plugins.js`, `plugins-dev.js`, `lint.js`, `bindings.js`) against the 2026-07-18 baseline, plus a 405 → 477 rule-count drift.

Both deserve their own fix; neither belongs in this diff.

## After merge

Nothing ships behaviourally — `detect-non-literal-fs-filename` reports exactly what it reported before. The value is that the next probe of this shape terminates at the header entry instead of at the sanitiser.


ofri-peretz:
### CI status, and a correction to the PR body

**Correction:** `Portability Audit` is **SUCCESS** in CI. The hash drift I hit at push time was local — this worktree shares the main checkout's `node_modules`, so it was hashing a different installed oxlint than CI resolves. The `portability` hook exclusion was unnecessary; the gate was never red for a real reason. `docs#test` / `tests-affected` stands as written (`@interlace/ui/effects/shimmer-button` is not an export on `origin/main` either).

**What is red, and why it is not this PR:**

| check | verdict |
| --- | --- |
| `Supply-chain floor` | FAILURE — OSV finds 13 known vulns across **committed lockfiles**: `fast-uri` 3.1.5 → 3.1.6 (4 CVEs, in root + `.github/deps/cyclonedx`), `qs` 6.15.3 → 6.16.0, `@humanfs/node` 0.16.7 → 0.16.8 (root + two benchmark lockfiles). |
| `Quality Gate` | FAILURE **only** because `R_SUPPLY: failure`. Every other input is green — `R_OXLINT`, `R_MARKDOWN`, `R_CHANGELOG`, `R_WORKFLOWS`, `R_TAXONOMY`, `R_PUBLISH`, `R_LOCKFILE`, `R_RECALL`, `R_BENCH_SIZE` all `success`. `R_RECALL` green matters here: the CWE corpus recall gate did not move. |

This diff is 3 files, +82 lines, **zero lockfiles touched** — it cannot have caused either. The `fast-uri` / `qs` / `@humanfs/node` bumps are a separate change, and per the working agreement they belong in their own intent rather than smuggled in here.

Not merging on my own: a `FAILURE` is explicitly not what `--admin` is for, even a pre-existing one.

PR #829, opened 2026-09-03
chore(ci): park the issue-sweep cron; the Routine owns the schedule now

ofri-peretz:
The scheduled issue sweep now runs as a Claude Code Routine (same cadence, `40 */6 * * *` UTC, same `chore/issue-sweep` branch, same one-open-PR guard). The Actions version could not edit `.github/workflows/`, which is how #823 landed a lock test without the two edits it locks. The Routine runs with credentials that can.

Two changes to `issue-sweep.yml`:
- the cron is commented out; `workflow_dispatch` stays as the fallback.
- a `pull_request` trigger arms squash auto-merge when a PR opens from `chore/issue-sweep`. The Routine never holds the merge tool, same split as before.

Merge after the Routine's first real sweep succeeds.


ofri-peretz:
Re-firing pull_request events: GitHub Actions created no check suites for a7ba769f2. Reopening immediately.

ofri-peretz:
History note: commit a7ba769f2 was made in a worktree whose checkout had been interrupted, so its tree dropped most of the repo. That is why the PR went DIRTY, GitHub ran no pull_request workflows on it, and Vercel reported "No Next.js version detected". 2fbbecf54 restores the full tree; the net diff of this PR against main is still the single issue-sweep.yml change (cron parked + auto-merge arming job). Squash-merge as usual.

PR #913, opened 2026-09-06
fix(benchmarks): @angular-eslint is measurable — let the lint name its own type-aware rules

### LINKED ISSUE #897: ILB-Arena: @angular-eslint has never produced a real measurement

ofri-peretz:
Closes #897.

## The problem

`@angular-eslint` has never produced a real measurement on **any** ESLint major. Its arena config enables all 48 rules; some need a TypeScript program the `.js` corpus cannot provide. Until run failures became fatal (#898) that surfaced as a silent **0/40 on every published run** — a fabricated number sitting in the comparison table next to real ones. #898 marked it `unmeasurable`, which was honest but still not a measurement.

Metadata cannot solve it:

```
total 50 | requiresTypeChecking 0
configs: []
```

Zero rules declare type-awareness, and the plugin exports no configs — so there is no recommended set to borrow and no static way to identify the offenders.

## The approach

ESLint names the rule in its error, so let the lint tell us: drop that rule, retry, repeat, bounded by the rule count. No hardcoded list, and a rule that stops requiring types starts being measured again on its own.

It converges in three passes and reports what it dropped:

```
ⓘ 3 rule(s) need a TypeScript program and were disabled for this corpus:
  @angular-eslint/no-developer-preview, @angular-eslint/no-experimental,
  @angular-eslint/no-uncalled-signals
```

Angular now runs **45 of 48 rules** and scores **0/40 TP, 0 FP**. That zero is *earned* — the rules executed and found nothing, which is the correct answer for an Angular template plugin on a security corpus — where the old zero was fabricated. The dropped rules print alongside so the number is interpretable rather than mysterious.

## My own lock caught the refactor

Moving the retry into `runEslint` pushed the fatal-failure decision one level out into `runEslintOuter`. The lock from #898 anchors on the catch after `lintFiles(`, which is now the *retry's* catch, so it failed immediately.

I re-pointed it at `runEslintOuter` rather than loosening it, and re-proved it: removing the fatal exit still fails it 1/12. The invariant it guards — a failed lint run is never scored as "found nothing" — is unchanged.

## Test plan

- [x] v10: 18 plugins scored, no "Not scored" entries, angular measured.
- [x] v8: same, and unicorn still excluded on its declared `>=10.4`.
- [x] `benchmarks`: 283 tests, 10 files, all pass.
- [x] Lock re-proven to bite after re-pointing.
- [x] Full pre-commit gate green.

### DIFF
--- a/benchmarks/__tests__/peer-eslint-support.lock.test.ts
+++ b/benchmarks/__tests__/peer-eslint-support.lock.test.ts
@@ -75,12 +75,18 @@ describe('a failed lint run is never scored as "found nothing"', () => {
     'utf-8',
   );
 
-  /** The body of runEslint's lint-run catch block. */
+  /*
+   * The catch that decides what a failed lint run means. It lives in
+   * runEslintOuter: runEslint retries a rule needing a TypeScript program
+   * (#897) and rethrows anything else, so the fatal decision sits one level
+   * out. Anchored on the function, not the lintFiles call, which is now
+   * inside the retry loop.
+   */
   const runCatch = (): string => {
-    const at = RUN.indexOf('const results = await eslint.lintFiles(');
+    const at = RUN.indexOf('async function runEslintOuter(');
     expect(
       at,
-      'lintFiles call not found — this lock is pointed at nothing',
+      'runEslintOuter not found — this lock is pointed at nothing',
     ).toBeGreaterThan(-1);
     const rest = RUN.slice(at);
     const start = rest.indexOf('} catch (e) {');
--- a/benchmarks/suites/ilb-arena/run.js
+++ b/benchmarks/suites/ilb-arena/run.js
@@ -206,18 +206,6 @@ const ALL_PLUGINS = [
     category: 'Vue.js',
   },
   {
-    /*
-     * Its recommended set includes rules that need a TypeScript program
-     * (@angular-eslint/no-developer-preview throws "requires type
-     * information"), and this corpus is plain .js with no project. The plugin
-     * does not mark those rules `requiresTypeChecking`, so they cannot be
-     * filtered out by metadata.
-     *
-     * This was always true. Until run failures became fatal it surfaced as a
-     * silent 0/40 — a fabricated measurement — on every run, v10 included.
-     * Declared, not scored, until the corpus can give it a program (#897).
-     */
-    unmeasurable: 'needs a TypeScript program; the arena corpus is plain .js',
     name: 'angular',
     displayName: '@angular-eslint/eslint-plugin',
     config: './configs/angular.config.js',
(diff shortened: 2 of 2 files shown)


ofri-peretz:
Superseded — #897 was fixed on main while this was in review, by excluding the three rules by name in `angular.config.js`. Closing rather than merging a competing rewrite of a landed fix.

We reached the same answer by different routes, which is a useful cross-check: they bisected all 50 rules against the corpus, I let ESLint name the offender and retried. **Both identified exactly the same three** — `no-developer-preview`, `no-experimental`, `no-uncalled-signals` — so the set is right.

Verified main's version locally:

```
18 plugins scored, exit 0
@angular-eslint/eslint-plugin → 0/40 TP
```

Angular is measured, and that 0 is earned rather than fabricated. #897 is genuinely closed.

## The one difference, for the record

Main's list is maintained by hand; mine derived the set at runtime, so a rule that starts or stops requiring a TypeScript program is picked up without an edit, and it generalises to any peer with the same problem.

I am **not** proposing to reopen that. The hardcoded list is simpler, it is already in, and three names that change roughly never is not a maintenance burden worth churn. Worth revisiting only if a second peer hits this or angular's type-aware set shifts — at which point the derive approach is in this branch's history (`cbdf91ef6`).

One thing from here that may be worth keeping independently: the lock re-pointing. Moving the retry changed where the fatal-failure decision lives, and the #898 lock caught it immediately — it anchors on the catch after `lintFiles(`. Main's fix does not move that code, so the lock is fine as-is today. Noting it only because the next person who restructures `runEslint` will trip the same wire, which is the lock working as intended.

PR #917, opened 2026-09-06
fix(benchmarks): angular is measurable, and a peer crash no longer blanks the matrix

### LINKED ISSUE #897: ILB-Arena: @angular-eslint has never produced a real measurement

ofri-peretz:
Closes #897.

## angular was never unmeasurable

It was declared *"needs a TypeScript program"* on the strength of one rule's error message. Bisecting **all 50 rules** against the corpus — running each alone, recording which threw — shows **47 run fine and exactly 3 need a program**:

```
no-developer-preview · no-experimental · no-uncalled-signals
```

They're excluded by name because they cannot be found any other way: `requiresTypeChecking` is set on **0 of 50** rules, so metadata filtering has nothing to filter. The config asserts the expected count, so an upstream addition fails at load rather than silently widening the exclusion.

**angular now scores a real 0/40, F1 0.0%** — a true measurement replacing a fabricated one, and it means something different. This corpus is Node and browser security fixtures; Angular's rules look for Angular TypeScript. A 0 says *the corpus holds nothing in its domain*, not that the plugin misses vulnerabilities. The plugin table says so, because the number alone doesn't.

## One peer's crash left the whole matrix unmeasured

#891 made a failed lint run fatal — right about the plugin that fails, wrong about the other seventeen.

`unicorn@72.0.0` declares `eslint: ">=10.4"`, we run `10.7.0`, and `unicorn/prefer-string-slice` still throws `'text' must be a string` on this corpus. So `process.exit(3)` fired and **the arena published nothing at all** — leaving the last published numbers, the ones full of the fabricated zeros #891 set out to kill, as the newest ones.

A crash is now its own state: recorded with its message, reported by name, never scored, never fatal for somebody else's bug. **Our own plugin still stops the run**, because that one is ours.

The run completes end to end for the first time:

```
exit 0 · 17 scored · unicorn named as crashed · Interlace 39/40 F1 94.0%
```

## The lock that caught me

`peer-eslint-support.lock` failed on this change, **correctly**. It asserted the *mechanism* — "contains `process.exit(3)`" — inside a fixed 1600-byte window, and the added branch pushed the exit past it.

Rewritten to assert the property in three parts: never return `[]` whoever failed; still exit when **our** plugin fails; record a crash for a peer. The window is now sized to the block's closing brace instead of a guessed byte count.

## Verification

- [x] Benchmarks suite: **102 failures before this change and 102 after**, same 4 files — every one a plugin-load failure in an unbuilt worktree. Measured both ways rather than assumed.
- [x] Whole-graph typecheck: **562 errors before and 562 after**, identical — the worktree resolves the devkit through another checkout's `node_modules`.
- [x] `configs-load` suite: 20/20 pass.
- [x] `tests-affected`, `typecheck` and `build` excluded for this push, for the reasons measured above — **not `--no-verify`**: `shim-drift`, `format-drift`, `stale-build-artifacts` and `commitlint` all ran and passed. CI builds the packages and runs the rest for real.

## Not included, deliberately

**No regenerated results.** The published files still hold zeros for plugins that do work — `eslint-plugin-security` records `0/0` there and scores **11/40, F1 37.3%** on a live run. Regeneration belongs to a proper fixture run, not to the patched-together local install I used to verify this.

## Two things worth their own issues

1. **`securityRelevant` is `true` for 17 of 18 peers** — including vue, jest, react, jsx-a11y and promise. It's opt-out and only jsdoc opted out, so the published ledger's `securityRelevantPlugins: 17` claims something that isn't so.
2. **unicorn's crash is an upstream defect** on a major it declares support for. Worth reporting to them; the arena now records it instead of dying on it.

### DIFF
--- a/.github/workflows/quality-full.yml
+++ b/.github/workflows/quality-full.yml
@@ -247,7 +247,13 @@ jobs:
           # change in this repo) skipped measurement entirely while the workflow
           # comment claimed feedback was merely "slightly delayed". Source is
           # exactly what moves the number, so source has to open the gate.
-          if echo "$CHANGED" | grep -qE '^(benchmarks/suites/.*\.config\.(js|mjs)|benchmarks/package\.json|benchmarks/__tests__/configs-load\.test\.ts|benchmarks/vitest\.configs-load\.mts|packages/[^/]+/package\.json|packages/[^/]+/src/.*|scripts/(check-artifact-size|devkit-infra-metrics)\.ts|\.agent/(artifact-size-baseline|devkit-infra-baseline)\.json|\.github/workflows/quality-full\.yml)$'; then
+          #
+          # `benchmarks/suites/**` and `benchmarks/__tests__/**` are whole
+          # directories rather than the two filenames that used to be listed.
+          # ILB-Arena's runner and its peer-support module sit beside the
+          # configs and are what the configs are loaded to exercise; a PR
+          # rewriting run.js matched nothing here and this job did not run.
+          if echo "$CHANGED" | grep -qE '^(benchmarks/suites/.*|benchmarks/package\.json|benchmarks/__tests__/.*|benchmarks/vitest\.configs-load\.mts|packages/[^/]+/package\.json|packages/[^/]+/src/.*|scripts/(check-artifact-size|devkit-infra-metrics)\.ts|\.agent/(artifact-size-baseline|devkit-infra-baseline)\.json|\.github/workflows/quality-full\.yml)$'; then
             echo "run=true" >> "$GITHUB_OUTPUT"
           else
             echo "run=false" >> "$GITHUB_OUTPUT"
@@ -387,6 +393,14 @@ jobs:
         run: npx turbo run build --filter='eslint-plugin-*' --filter=@interlace/eslint-devkit
       - name: Load every benchmark config
         run: npm run --silent -w @interlace/benchmarks test:configs-load
+      # The rest of `benchmarks/__tests__` runs on the unit lanes. This one is
+      # pulled forward because it guards the arena, and the lanes are skipped
+      # for exactly the PRs that touch the arena: their gate opens on
+      # `packages/*/src`, which an arena-only change never touches. A lock that
+      # never runs on the PRs that can break it is the "green locally, never
+      # run by the gate" failure recorded above scripts-tests.
+      - name: ILB-Arena peer-support locks
+        run: npx vitest run --root benchmarks __tests__/peer-eslint-support.lock.test.ts
       # Same job because it needs the same thing this one already paid for: the
       # built dist/. The contradiction only exists in the RENDERED message, and
       # the score in it is enriched from the rule's CWE — a source-only checker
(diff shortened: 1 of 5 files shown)


ofri-peretz:
Both findings addressed in c50c27a03.

**Finding 1 — the assertion was vacuous.** Correct, and worse than described: `runnable` is *built* by filtering `NEEDS_TYPE_PROGRAM` out of `angular.rules`, so `runnable.length` and `expected` are the same subtraction written twice. It could only ever catch a stale entry.

Replaced with two assertions, because neither implies the other:

- `MEASURED_RULE_TOTAL = 50` pinned to what the per-rule bisect ran against — this is what makes "upstream added a rule" observable at all;
- every name in `NEEDS_TYPE_PROGRAM` must still be published — a rename holds the total at 50 while leaving a live type-aware rule enabled.

While doing this I found the bisect had run against **22.1.0** (stale local `node_modules`) while the lockfile installs **22.2.0**. Checked 22.2.0 directly: identical rule names, same 50, `requiresTypeChecking` on none of them. So the pin is right for the version CI actually installs — but the pin is the reason that discrepancy would have surfaced at all next time.

**Finding 2 — no lock on the guard.** Added, and stronger than a `toContain`: alongside asserting the pinned total, the stale-name check, and that the cancelling comparison cannot return, there is a pure reproduction that runs both formulas against a grown rule set — old one passes, new one throws. That is the defect this PR fixed, executable.

**Non-blocking observation on load-catch vs run-catch** — taken; the config-load catch now says why it stays fatal for every plugin: nothing at that point distinguishes a broken plugin from a broken bench, unlike a lint-time throw, which is one upstream rule misbehaving on this corpus.

`crashMessage` and falsy messages: left as-is, `crash()` always passes a non-empty string.

Verified: `test:configs-load` 20/20; lock suite 18 pass. One pre-existing failure in that file locally (`peerEslintRange('unicorn')` reads `>=9.38.0`) is a stale `benchmarks/node_modules/eslint-plugin-unicorn@62` in this worktree against a lockfile pinning 74 — `origin/main`'s own copy of the test fails the same way here, two tests instead of one.

Issue #916, opened 2026-09-06
Release: @interlace/eslint-formatter-sarif@0.2.0 cannot be published — npm token cannot create a new package


ofri-peretz:
#909 merged and the release workflow **failed** at the publish step. [Run 34064489449](https://github.com/ofri-peretz/eslint/actions/runs/34064489449).

Everything upstream succeeded — detect, build, dist-integrity on eslint 8/9/10, and provenance was even signed and logged to Sigstore. It failed on the registry write:

```
npm notice publish Signed provenance statement with source and build information from GitHub Actions
npm notice publish Provenance statement published to transparency log: .../logIndex=2743936999
npm error code E404
npm error 404 Not Found - PUT https://registry.npmjs.org/@interlace%2feslint-formatter-sarif
npm error 404  The requested resource '@interlace/eslint-formatter-sarif@0.2.0'
              could not be found or you do not have permission to access it.
```

## Diagnosis

E404 on a **PUT** to a package that has never existed is npm's response for "you may not create this". The scope is fine and the token clearly works for existing packages:

| package | npm |
|---|---|
| `@interlace/eslint-devkit` | 1.19.2 ✅ |
| `@interlace/eslint-formatter` | not on npm |
| `@interlace/eslint-formatter-sarif` | **not on npm** |

So this is not a scope problem and not a code problem. The signature is a **granular access token scoped to selected packages** — it can publish updates to the packages on its list, and 404s on a name that is not on it, because the name does not exist to that token.

`@interlace/eslint-devkit` is the only `@interlace/*` package ever published, which is why this has not been hit before: #900 made a second one publishable for the first time.

## Fix — needs npm account access, so not something I can do

Any one of these:

1. **Add the package to the token's allowlist.** Requires the package to exist, so this pairs with (2), or with switching the token to *All packages* first.
2. **Publish 0.2.0 once by hand** to create it, then automation takes over:
   ```bash
   npm publish --workspace @interlace/eslint-formatter-sarif --access public
   ```
   Note this forfeits provenance for that one version unless run from CI.
3. **Widen the automation token** to *All packages* (or add the package after creating it), then re-run the failed job — which keeps provenance.

(3) then re-running the workflow is the cleanest: it preserves the SLSA attestation that every other package has.

## State right now

- `main` has `@interlace/eslint-formatter-sarif@0.2.0` in its manifest and a git tag.
- npm has nothing.

[...]

PR #936, opened 2026-09-07
fix(eslint-formatter): ship the schema its exports declare

ofri-peretz:
The 0.2.0 release never reached npm, and **not for the reason I predicted**.

I expected the npm-token wall from [#916](https://github.com/ofri-peretz/eslint/issues/916). It failed earlier than that — Dist integrity failed, cancelling its siblings and skipping the publish job entirely, so neither package shipped. `eslint-plugin-import-next` is still at 2.7.5.

The gate was right:

```
❌ 1 declared export path(s) missing from the tarball:
   @interlace/eslint-formatter  "./schema.json" -> schema.json  (NOT SHIPPED)
   A consumer importing that subpath gets "Cannot find module".

❌ 1 package(s) missing discoverability metadata:
   @interlace/eslint-formatter  funding
```

Both defects are mine, and both are only observable once the package is actually **packed** — the same reason `private: true` hid its missing build script and its broken emit layout in #928.

## The cause

`build-package.ts` publishes FROM `dist/` and copied a fixed asset list — `README.md`, `LICENSE`, `.npmignore`. A root asset a package declares in `files` never made the tarball. `schema.json` is in this package's `files` **and** mapped by `exports`, and still did not ship.

It now also copies any root-level file the manifest says it ships, skipping directories since `src/` and `dist/` are the build's own output. **That fixes the class rather than the instance** — the next package to declare a root asset gets it shipped without touching this script.

`funding` added, matching `eslint-formatter-sarif` and `eslint-devkit`.

## Test plan

- [x] Verified by **packing**, not reading: `tar tzf` now lists `package/schema.json`.
- [x] Published-artifact gate passes across **all 33 packages** — every declared export resolves, metadata complete, 3,613.8 kB unpacked.
- [x] Full build of all 33 packages clean.
- [x] Patch changeset included.

## After merge

The release retries. It will then reach the step I originally expected and stop there: the npm token cannot create a new package under the `@interlace` scope, which now blocks **two** packages (#916). One token fix ships both.

### DIFF
--- a/.changeset/formatter-ships-its-schema.md
+++ b/.changeset/formatter-ships-its-schema.md
@@ -1,0 +1,5 @@
+---
+'@interlace/eslint-formatter': patch
+---
+
+fix: `./schema.json` now reaches the published tarball, and the package declares `funding`
--- a/packages/eslint-formatter/package.json
+++ b/packages/eslint-formatter/package.json
@@ -16,6 +16,10 @@
   },
   "author": "Ofri Peretz <ofriperetzdev@gmail.com>",
   "license": "MIT",
+  "funding": {
+    "type": "github",
+    "url": "https://github.com/ofri-peretz/eslint"
+  },
   "homepage": "https://github.com/ofri-peretz/eslint/tree/main/packages/eslint-formatter#readme",
   "repository": {
     "type": "git",
--- a/scripts/build-package.ts
+++ b/scripts/build-package.ts
@@ -24,6 +24,7 @@ import {
   existsSync,
   mkdirSync,
   copyFileSync,
+  statSync,
   readFileSync,
   readdirSync,
   rmSync,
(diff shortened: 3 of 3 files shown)


ofri-peretz:
Superseded by #935, which landed the same two fixes while this was in review — and did the build change **better than mine**.

We hit the same failure and reached the same diagnosis independently. The difference is one I got wrong:

My version copied every root-level file named in `files`. `@interlace/eslint-formatter` lists `CHANGELOG.md` in `files`, so mine would have started shipping changelogs again — 225 kB across the ecosystem that `build-package.ts` deliberately excludes, with the rationale documented a few lines above the code I was editing. I reintroduced the thing the surrounding comment exists to prevent.

#935 has `EXCLUDED_ASSETS = new Set(['CHANGELOG.md'])` and keeps that exclusion explicit. Its framing is also sharper: the hardcoded list was a *third* hand-maintained inventory beside `files` and `exports`, and the three had drifted.

Verified main end to end by packing, not reading:

```
$ tar tzf interlace-eslint-formatter-*.tgz | grep -E 'schema|CHANGELOG'
package/schema.json
```

`schema.json` ships, `CHANGELOG.md` does not, and `funding` is present on main. Both defects from the failed 0.2.0 release are fixed there, so there is nothing left in this branch worth carrying.

The release can retry. It will now get past Dist integrity and stop at the step I originally expected — the npm token cannot create a new package under the `@interlace` scope (#916), which blocks both formatters.

PR #960, opened 2026-09-09
feat(supabase-security): the SDK we use ourselves, and did not lint

### LINKED ISSUE #944: eslint-plugin-supabase-security: the SDK we use ourselves and do not lint

ofri-peretz:
Closes #944.

`@supabase/supabase-js` is a direct dependency of our own blog, used in five files, and we ship domain plugins for **nine databases and five AI SDKs** without one for it. Four rules, all AST-structural.

## The rule that justifies the package

**`no-service-role-key-in-client`** — CWE-798, CVSS 9.8.

`service_role` bypasses Row Level Security completely; it is a database superuser in a string. The reason this needs a linter rather than a code review is that the line looks like every other one:

```ts
const supabase = createClient(url, key);
```

Nothing at the call site distinguishes the two keys. Two independent shapes are reported:

1. A service-role variable behind a prefix whose entire purpose is to inline into the browser bundle — `NEXT_PUBLIC_`, `VITE_`, and five more.
2. A service-role variable read inside a `"use client"` module.

It abstains on `server-only`, where the read is correct and the bundler already enforces it.

## The other three

| rule | why | preset |
|---|---|---|
| `no-dynamic-rpc-name` | a computed `.rpc()` name lets the caller pick which Postgres function runs — next to `delete_user` in the same exposed schema | recommended |
| `require-auth-error-check` | `getUser()` reports a forged token by **returning** `error`, not throwing, so `const { data } =` reads an auth failure as an anonymous success | recommended |
| `no-public-storage-bucket` | a public bucket serves every object to anyone with the URL | **strict only** |

`no-public-storage-bucket` waits in `strict` because a bucket of site assets is legitimately public and the source cannot tell that from a bucket of invoices.

## What it deliberately does not do

The issue's second suggestion — `.from()` with no filter, as a proxy for "RLS assumed rather than enforced". Whether a policy covers a query lives in the database, not the source. That rule would fire on every admin query and every count. A rule that cannot be right is worse than no rule.

## The repo's gates found two things I had wrong

Which is the argument for having them:

- The **spelling audit** caught `envVarName` reading `.property.name` directly, so `process['env'].X` would have walked straight past — a blind spot the person writing the leak does not share. Rewritten on `propertyName`, with a case pinning it.
- The **coverage gate** caught four branches my tests never reached, including one only a `for…of` head can produce, since `let { data };` is a syntax error.

## Test plan

- [x] **105 tests, 100%** statements / branches / functions / lines.
- [x] Every suite carries the repo's litmus explicitly — rename every identifier to `foo`/`bar` and the rule still fires, because it keys on call shapes and on the property read off `process.env`, never on a name.
- [x] `module-gate.lock.test.ts` holds the `@supabase/*` gate over the **whole registry** rather than per rule, so a rule added later is covered the day it lands. Its negative fixtures are an ORM that owns `.rpc` and `.auth`, a store that owns `createBucket`, and a config naming a service-role variable for something else — precisely the code this plugin would report on if the gate ever came off.
- [x] Dogfooded on a file carrying all four defects; all four report with CWE, OWASP and CVSS.
- [x] `turbo typecheck test` **90/90**, **2098** docs tests, oxlint shims in sync at **481** rules.

## One change outside the package

The new oxlint shim is generated, and the generator emitted two constructs prettier rewrites (`function(`, a long `require(...)`). The format baseline is **shrink-only** by its own note, so the shim could not simply join it — the fix was to make the template emit what prettier already wants. All 31 shims are now clean and **30 of them left the baseline**: 3612 → 3550, with **nothing joining** (verified by diffing the two lists, not by trusting the count).

## After merge

The root README stays unchanged on purpose — it lists *published* plugins, and this is `0.1.0` until the release PR lands.


ofri-peretz:
Three CI findings addressed, all of them real:

**Plugin Taxonomy — the best of the three.** Four `PUBLIC_PREFIXES` entries had no case that would fail if the row were deleted: `REACT_APP_`, `GATSBY_`, `NUXT_PUBLIC_`, `EXPO_PUBLIC_`. A prefix nothing exercises is a prefix nobody knows is broken. Added one case per row, and `PUBLIC_` (SvelteKit) last on purpose — every other entry contains it as a substring, so a case using bare `PUBLIC_` would have passed for the wrong reason. That is exactly the trap the gate's own message warns about (`totpSecret` covers `secret`, not `totp`).

**Publish Metadata / Changeset present.** The changeset covered only the plugin, but this PR also touches `@interlace/eslint-devkit` — `PLUGIN_DOCS_CATEGORY`. Without that entry every rule ships with no `meta.docs.url`, which is a "see docs" link that 404s in every IDE, CI report and SARIF file. Added the devkit as a `patch` with that reason recorded, rather than reaching for `skip-changeset`: it is not internal-only, it is the difference between working and broken doc links.

**Lockfile Sync.** My local `npm install` had written an `"extraneous": true` copy of prettier under `@posthog/cli` — junk from this machine, not from `package.json`. Reset the lockfile to `main` and re-resolved with the check's own command, then confirmed it is idempotent: running `npm install --package-lock-only --ignore-scripts --no-audit --no-fund` again produces no further change. The remaining diff is the two `eslint-plugin-supabase-security` entries plus a `smol-toml` removal that npm's own resolution wants.

110 tests, still 100% on all four counters.

PR #979, opened 2026-09-13

ofri-peretz:
## Summary

- **#973 — Release pipeline is stalled**: Fixed. The scheduled `Release Liveness` cron is not guaranteed to fire on its six-hour boundary (GitHub documents scheduled triggers as best-effort and delays them under load), and on 2026-09-10 one fired ~3h49m late, landing inside a still-finishing release run. `eslint-plugin-maintainability`'s publish job had already exited 0 two minutes earlier, but `npm view eslint-plugin-maintainability@latest` still answered the prior version, and the check filed "Release pipeline is stalled" against a pipeline that was healthy — confirmed live against the npm registry (`https://registry.npmjs.org/eslint-plugin-maintainability` now reports `latest: 3.2.6`, matching `main`).

  `--skip-publish-lag` already covers the one race this script's author had timed (the `workflow_run` trigger firing on the release commit itself). It does nothing for a `schedule` or `workflow_dispatch` run that happens to overlap someone else's in-flight release, which is what #973 was.

  Added a 20-minute publish grace period to `scripts/check-release-liveness.ts`, keyed off how long ago the package's version actually changed on `main` (via git commit history), so a bump that is still fresh is not treated as a stall regardless of which trigger fired the run. 20 minutes is roughly double the ~9.5 minutes the #973 release run took end to end, so a genuine multi-hour stall is still caught on the next six-hourly run.

  - Lock: `scripts/__tests__/check-release-liveness.test.ts` — "a very recent bump gets a publish grace period (#973)" (a bump seconds old is not flagged; the same bump backdated past the grace window still is). The pre-existing "flags a package whose version on main is ahead of npm" test, and the shared `beforeEach` baseline commit, were backdated so they keep testing a genuine stall rather than accidentally landing inside the new grace window.

## Deliberately NOT touched

- **#916 — `@interlace/eslint-formatter-sarif@0.2.0` cannot be published (npm token cannot create a new package)**: Confirmed via the failed run's logs (`npm error code E404 ... PUT .../@interlace%2feslint-formatter-sarif — Not found`) — the granular npm access token is scoped to an allowlist of existing packages and 404s on any package name it doesn't recognize, which is npm's behavior for "you may not create this." There is no code change in this repository that can widen an npm access token's package allowlist or publish a package for the first time by hand; the issue's own body says as much ("not something I can do"). It needs a human with npm account access to either add the package to the token's allowlist, publish `0.2.0` once by hand, or widen the token to *All packages*. The same root cause affects `@interlace/eslint-formatter` per the issue body.
- **#764 — CI: CodeQL analyses nightly only, not per-PR and no longer post-merge**: Not a bug report. It is a tracking issue for an already-accepted, documented trade-off (`docs/adr/0001-codeql-runs-post-merge-not-per-pr.md`, intent `docs/intents/ci-speed/`), listing the conditions under which to *revisit* the decision (a CodeQL finding reaching `main` and causing harm, the dropped Scorecard SAST ratio costing more than the runner slots it bought, faster incremental analysis from GitHub, or the repo starting to release from PR branches). None of those conditions are asserted as true in the issue itself, so there is nothing to fix — reverting the trigger would undo a deliberate decision, not correct a defect.

## Test plan

- [x] `npx vitest run scripts/__tests__/check-release-liveness.test.ts scripts/__tests__/release-liveness-race-lock.test.ts` — 17/17 passed.
- [x] Confirmed live: `https://registry.npmjs.org/eslint-plugin-maintainability` dist-tag `latest` is `3.2.6`, matching `main` — the incident that opened #973 was a false alarm, not an actual stall.
- [x] New lock: `scripts/__tests__/check-release-liveness.test.ts` › "a very recent bump gets a publish grace period (#973)".

## After merge

`release-liveness.yml`'s six-hourly cron and `workflow_dispatch` runs are less likely to false-positive when they happen to land during an active release. A genuine multi-hour stall is still caught on the next six-hourly run. #916 and #764 remain open — #916 pending a human with npm account access, #764 as an informational tracker — do not close either from this PR.

https://claude.ai/code/session_0115oUTeFkv3xM6MyKbantsJ

---
_Generated by [Claude Code](https://claude.ai/code/session_0115oUTeFkv3xM6MyKbantsJ)_


ofri-peretz:
Closing this one — it's redundant. Between this PR's branch being cut from `main` and this PR being opened, a concurrent run of the same issue-sweep routine independently reached the same diagnosis for #973 (npm registry read-after-write propagation lag racing the scheduled check), opened #978 from the same `chore/issue-sweep` branch name, and it was auto-merged by this repo's own `issue-sweep.yml` `arm` job within ~2 minutes — before this PR existed. That's why this PR now shows a merge conflict against `main`: `main` already carries an equivalent fix (a retry-based one; this PR's was a grace-period-based one) touching the same lines of `scripts/check-release-liveness.ts`.

Stacking a second guard for the same failure mode on top of #978's would be unnecessary duplication, not an improvement, so I'm closing this rather than resolving the conflict. #916 and #764 are unaffected by either PR and remain open for the reasons stated in #978's description (npm token needs human access; #764 is a tracking issue for an accepted trade-off, not a bug).

Worth a look separately: the two-session race itself (chore/issue-sweep branch, two nearly-simultaneous invocations of this sweep). The routine's own "refuse to start while a PR is already open" guard doesn't close the TOCTOU window between two runs that both check before either has opened a PR.

---

PR #1008, opened 2026-09-14
fix(secure-coding): the holder-property exemption no longer survives a replacement

### LINKED ISSUE #998: detect-object-injection: the holder-property exemption survives a replacement (FN)
Raised by CodeRabbit on #997 and **confirmed by reproduction**, not just by reading.

#997 added an exemption for a null-prototype map held as a property of a holder object (`const flags = { bools: Object.create(null) }`). It resolves the holder, finds the property in the *initializer*, and exempts the indexed write.

The initializer says what the property was **created** as. It does not say what it **is** at the write.

## Reproduction

Run against the rule as it stands on `main`. All three are silent; B and C should report.

```ts
// A — CONTROL, correctly exempt
const flags = { bools: Object.create(null) };
export function f(req: any) { const k = req.query.k; flags.bools[k] = 1; }

// B — the property is replaced with an ordinary object
const flags = { bools: Object.create(null) };
flags.bools = {};
export function f(req: any) { const k = req.query.k; flags.bools[k] = 1; }

// C — a spread rebuild replaces it
let flags = { bools: Object.create(null) };
flags = { ...flags, bools: {} };
export function f(req: any) { const k = req.query.k; flags.bools[k] = 1; }
```

B and C write an attacker-controlled key into a **prototype-bearing** object and the rule stays silent. That is a false negative in the rule whose entire purpose is prototype pollution — the direction that matters most here.

## Why this is not a quick fix

I tried the conservative narrowing — withdraw the exemption if the holder binding or that property is written anywhere in the file — and it broke five previously-passing exempt cases, including the burgee one (`packages/burgee/src/yargs-parser.ts:188`) that motivated the exemption. Proving *which* value reaches the write needs flow analysis the rule does not currently have, so a correct fix is real work rather than a guard clause. Reverted rather than shipped half-done.

## Options

1. **Narrow the exemption properly** — track writes between the declaration and the indexed access. Correct, and the most work.
2. **Require the holder to be `const` and the property never assigned** — cheaper, still needs a write-scan that does not misfire on the five exempt cases.
3. **Withdraw the holder-property exemption** and take the false positives back. Recall-safe, and the repo's own stance is that "a recall loss is the worse trade for a security rule".

Recorded against a reproduction so the choice is made against evidence rather than an argument.

ofri-peretz:
## Summary

**Closes #998.**

The exemption for a null-prototype map held on a holder object reads the **initializer**. That says what the property was *created* as, not what it *is* at the indexed write. A later `flags.bools = {}` — or a `flags = { ...flags, bools: {} }` rebuild — swaps in an ordinary object, and honouring the declaration then suppressed a real finding on a prototype-**bearing** target.

A false negative, in the rule whose entire subject is prototype pollution.

```ts
const flags: any = { bools: Object.create(null) };
flags.bools = {};                                   // now an ordinary object
export function f(req: any) {
  const k = req.query.k;
  flags.bools[k] = 1;                               // was SILENT — now reports
}
```

## Approach

Proving which value reaches the write needs flow analysis this rule does not have, so the answer is conservative: if the holder binding or that exact property is assigned anywhere, the exemption is withdrawn. A recall loss is the worse trade for a security rule — the repo's own stated stance.

Three details decide whether it works at all:

- The write being judged is `holder.key[computed]`, whose own `propertyName` is the computed key rather than `key` — so the scan never matches the very write it is deciding about.
- **The walk skips `parent`.** Following it is a cycle, not a traversal, and blows the stack. That is exactly how my first attempt at this failed, and why it looked like the heuristic was too aggressive when it had in fact never executed.
- **No visited-set, no non-Identifier holder branch.** With `parent` skipped what remains is a tree, and `findInitializer` resolves only a bare local binding — so both would have been unreachable. Removed rather than left as uncovered lines.

## Test plan

- [x] Three new cases, **verified silent before the fix**: property reassigned, holder rebuilt via spread, and a nested holder that gets no exemption
- [x] All five previously-exempt cases still pass — including the burgee one (`packages/burgee/src/yargs-parser.ts:188`) that motivated the exemption in the first place
- [x] **3792/3792** across `eslint-plugin-secure-coding`
- [x] Coverage back to **100%** (the gate rejected the first attempt at 99.96%, which is what surfaced the two dead branches)


ofri-peretz:
Closing in favour of #1009, which landed the same fix 73 seconds later and is the more complete version.

Both close #998 and both take the same approach — withdraw the exemption when the holder binding or that property is assigned, because proving which value reaches the write needs flow analysis this rule does not have. #1009 additionally carries a changeset, which a rule-behaviour change needs and this PR was missing, and puts the regression cases in their own file rather than appending to the main suite.

One finding from here worth carrying over if it is not already in #1009: the tree walk must skip `parent`. Following it is a cycle rather than a traversal and blows the stack — that is what made my first attempt look like an over-aggressive heuristic when it had in fact never executed. The 100% coverage gate also rejects a visited-set and a non-Identifier holder branch as dead code, since `findInitializer` only ever resolves a bare local binding.

No duplicate merge, no conflict between the two.

ofri-peretz:
I built the same fix in parallel as #1009 before spotting this one — yours is first and I'm closing mine. Rather than waste the work, I ran **my** test file against **your** implementation as an adversarial check. It passes 7 of 8.

Passing on your branch, and worth folding in as extra coverage since you already satisfy them:

```
✓ a logical-assignment operator installs it just as plainly   flags.bools ??= {}
✓ a quoted replacement reaches the same property              flags['bools'] = {}
✓ replacing a sibling property leaves this one exempt         flags.strings = {}
✓ many writes through the property are still writes through it
```

That third one is the guard against over-narrowing (a sibling being replaced says nothing about this property), and the fourth is the one that broke the reverted attempt — two writes through `flags.bools[key]` must not read as a replacement. Both green on your branch; they'd just pin it.

The one that differs:

```
× deleting it means what is there afterwards is not the map    delete flags.bools
```

**I think your behaviour is right and my case was wrong**, which is why I'm raising it rather than filing it. After `delete flags.bools`, the property is `undefined`, so `flags.bools[k] = 1` throws a TypeError — it cannot pollute anything. The exemption's *premise* stops holding (the initializer no longer describes the property) but its *conclusion* still does, for a different reason. Reporting there would be a false positive dressed as a catch.

Worth a one-line comment either way, since the next person to read the guard will ask the same question.

The cases are in `holder-property-replacement.test.ts` on `fix/holder-property-replacement` if you want to lift them; otherwise ignore and I'll delete the branch with the PR.

PR #1009, opened 2026-09-14
fix(secure-coding): a replaced holder property is not exempt

### LINKED ISSUE #998: detect-object-injection: the holder-property exemption survives a replacement (FN)

ofri-peretz:
Closes #998.

## The false negative

[#997](https://github.com/ofri-peretz/eslint/pull/997) exempted a null-prototype map held as a property of a holder object, by resolving the holder and reading the property out of its initializer. The safety property is real — a null-prototype target has no prototype to pollute.

But an initializer says what the property **was** when the object was built. It says nothing about what it **is** at the indexed write:

```ts
const flags = { bools: Object.create(null) };
flags.bools = {};
flags.bools[req.query.k] = 1;   // silent
```

An attacker-controlled key landing in a prototype-bearing object, in the rule whose entire subject is prototype pollution. This repo's own stance is that a recall loss is the worse trade.

## Option 2, and why the first attempt had to be reverted

The issue records that "withdraw if the holder or the property is written anywhere in the file" broke five correctly-exempt cases. That failure is the fix: match on the assignment **target's shape**, not on "is this path ever written". It separates the two shapes with no flow analysis at all.

| | |
|---|---|
| `flags = …` | withdraw — the holder is rebound (covers the `{ ...flags }` rebuild) |
| `flags.bools = …` | withdraw — the property is replaced |
| `flags.bools[k] = …` | **keep** — a write *through* the map is the access being judged |

The reverted scan counted that last line as a replacement, which is exactly why it pulled the exemption from five exempt cases including the burgee one (`packages/burgee/src/yargs-parser.ts:188`) that motivated #997. Those five pass here untouched.

Also withdrawn by `??=` / `||=` and by `delete` — each installs or leaves something that is not the map — and matched through `propertyName`, so `flags['bools'] = {}` resolves the same as the dotted form.

## My first reproduction was wrong

Worth recording. A standalone file with `const k = req.query.k` stayed silent **even with a plain `{}` holder** — so the "silence" I measured was the rule not firing on that shape at all, not the exemption. A control caught it. Redone in the rule's own harness, which is authoritative.

## Test plan

- [x] New `holder-property-replacement.test.ts` — 3 valid, 5 invalid. On **unfixed** code all five invalid cases fail and the three valid ones already pass:

  ```
  ✓ A — the null-prototype map is still the null-prototype map
  ✓ many writes through the property are still writes through it
  ✓ replacing a sibling property leaves this one exempt
  × B — the property is replaced with an ordinary object
  × C — a spread rebuild replaces the holder binding
  × a logical-assignment operator installs it just as plainly
  × a quoted replacement reaches the same property
  × deleting it means what is there afterwards is not the map
  ```

  With the fix: 8/8.
- [x] `detect-object-injection` suite: **402 tests across 14 files**, all passing — the five exempt cases intact.
- [x] Whole plugin: **87 files, 3796 tests, 100%** statements / branches / functions / lines.


ofri-peretz:
Superseded by #1008, which is the same fix and was opened first. Closing rather than shipping two implementations of one guard.

I ran this PR's test file against #1008's implementation as an adversarial check before closing: **7 of 8 pass**, and the eighth (`delete flags.bools`) is a case where I now think #1008's behaviour is correct and mine was wrong — after the delete the write throws a TypeError rather than polluting anything. Details and the four extra cases worth lifting are on #1008.

ofri-peretz:
Closing again — the fix landed in #1012, which is merged, and #998 is closed.

Recording what happened, because it nearly cost the fix. #1008 and #1009 were the same fix built in parallel, and **each was closed in favour of the other**: #1008 said "closing in favour of #1009, the more complete version"; I closed #1009 in favour of #1008. For a while #998 had two good fixes and neither was open. I reopened this one on noticing, then found #1012 had already merged the work.

No action needed. The merged implementation is sound — I ran this PR's eight adversarial cases against it and it passes seven, the eighth (`delete flags.bools`) being one where I now think its behaviour is correct and my case was wrong: after the delete the write throws a TypeError rather than polluting anything.

The four extra cases worth lifting are listed on #1008 — `??=`, a quoted `flags['bools'] = {}`, a replaced *sibling* property staying exempt, and two writes through `flags.bools[key]` not reading as a replacement. All four are green against what merged, so they would pin behaviour rather than change it.

PR #1027, opened 2026-09-14
chore(deps): verify oxlint 1.82.0 and pin its new rolldown-runtime bundle

ofri-peretz:
## Summary

`main` is red on the portability audit. #1024 landed `oxlint` 1.81 → 1.82 on the three required checks; `Audit + shim drift (no build required)` is not required, so its failure went in with it. This is the fix.

The gate reports four violations: the installed version is past `VERIFIED_OXLINT_RANGE`, and `plugins-dev.js`, `lint.js` and `bindings.js` all drifted from their pins.

Verified the way the gate's own message prescribes, not by re-baselining:

- All 33 probes in `verify-oxlint-runtime.ts` pass on 1.82.0.
- The 1.81.0 tarball hashes **identically to all four current pins**, so the pins were correct and the drift is genuinely 1.81 → 1.82.
- A `;`-split diff of 1.81.0 against 1.82.0 accounts for every differing chunk:

| bundle | chunks | what they are |
| --- | --- | --- |
| `plugins.js` | **0** | byte-identical — the production plugin API surface, the bundle the blocker patterns actually read, did not move |
| `plugins-dev.js` | 6 | 4 are the rolldown prelude moving out to `rolldown-runtime.js`; 1 is a De Morgan rewrite in `isSerializable` (`!(typeof prop != "object" \|\| !prop)` → `typeof prop == "object" && prop` — the same predicate) |
| `lint.js` | 57 | 55 the same prelude extraction, 2 the version literal |
| `bindings.js` | 60 | 54 the version literal inside napi guards; the rest hoist `readFileSync` into the top destructure and drop a dead `new URL(".", import.meta.url).pathname` |

## Two judgment calls

**`rolldown-runtime.js` is added to the pinned set.** It is new in 1.82.0, and it holds exactly the prelude that moved *out* of `plugins-dev.js`. Pinning only the original four would have left the gate showing four green hashes while covering less than it did at 1.81.0.

**`min` stays `1.74.0`.** `verify-oxlint-runtime.ts --update` narrows it to the installed version, but the field is never read anywhere in the script and its own comments define it as the *oldest* version still verified. `--update` also deletes the accumulated 1.77/1.79/1.80/1.81 verification notes, which the file keeps on purpose — so the constant block is hand-written here, with the 1.82.0 note prepended and the history intact.

## Test plan

- [x] `npx tsx scripts/verify-oxlint-runtime.ts` — 33/33 probes pass on 1.82.0.
- [x] **RED before the fix**: `main`'s copy of `scripts/audit-rule-portability.ts` against the installed 1.82.0 → `✗ 4 threshold violation(s)`.
- [x] **GREEN after**: `npm run audit:portability:ci` → `✅ All thresholds satisfied`.
- [x] Committed through lefthook; no `--no-verify`.

## Summary by CodeRabbit

* **Chores**
  * Updated the verified Oxlint version range to 1.82.x.
  * Refreshed runtime verification data for the new version, including an additional runtime component.
  * Expanded verification notes to document compatibility and behavior checks for the 1.82.0 release.


ofri-peretz:
Following up on the `rolldown-runtime.js` finding — it was right, and #1028 fixes it.

Pulling on it found two more defects of the same shape in `--update`, all three of which are why this PR's block had to be hand-written:

1. The bundle list was hard-coded in the verifier, so the new fifth pin would have been deleted on the next bump. It is now **read** from what the audit already pins — zero parsed is a hard exit, not a silent narrowing. Hashing all of `dist` was considered and rejected: the pinned set is curated, and `cli.js` / `workspace.js` / `index.js` would have made the gate fire on oxlint churn this audit deliberately ignores.
2. `min` was rewritten to the installed version. It is the *oldest* version still verified and is read nowhere in the audit, so narrowing it discarded the record it exists to keep. Only `maxKnown` moves now.
3. The comment body — the actual verification record — was deleted wholesale. It now carries over verbatim.

`--update` against `main`'s committed block is now a no-op: range, all 5 pins and 74 comment lines preserved, byte-identical after prettier.

PR #1035, opened 2026-09-14
chore(benchmarks): revalidate the per-rule p95 budget against ILB-Flagship 2026-09-07

ofri-peretz:
## Summary

Closes the other half of #1032 (the first half landed in #1034). `benchmarks/budgets/per-rule-p95.json` read `lastValidated: 2026-05-09` — **128 days against a 90-day TTL**.

It was stale for a bookkeeping reason, not a measurement one. Flagship runs happened on 2026-08-25, 2026-08-31 and 2026-09-07 — the field was simply never updated afterwards, which is precisely what the artifact's own refresh note asks for: *"Update lastValidated after the next ilb:flagship run."*

## This is a revalidation, not a date bump

`npm run check:per-rule-budget` reads the latest `benchmarks/results/ilb-flagship/<date>.json` and reports:

```
Measurements: ILB-Flagship 2026-09-07
9/10 rules within budget · 0 over budget · 1 unmeasured
```

So `lastValidated` becomes **2026-09-07** — the date of the run that validated it — not today. `source` now points at the results file that run actually wrote, carrying the toolchain that file itself records (`eslintVersion`, `oxlintVersion`, `nodeVersion`): ESLint 9.39.4 + oxlint 1.63.0 + Node **24.20.0**. The previous string claimed Node 24.13.0, which no run reports.

Freshness gate after the change:

```
✅ Per-rule p95 budget: fresh (age 7d, TTL 90d)
```

## What I deliberately did not do

**Run `ilb:flagship` locally.** Timings from a laptop are not comparable to the CI runners these budgets gate against — that is exactly how the corpus-findings budget went wrong earlier this month (measured here, 6,106 findings; measured in CI, 6,238). The CI run of 2026-09-07 is the measurement; this PR records that it still holds.

## Still open, not hidden by this

`react-a11y/alt-text` has no measurement in the flagship JSON, so it cannot gate. It needs a corpus cell — benchmark authoring, not a refresh. The generated report names it explicitly rather than rounding 9/10 up to "green".

## Test plan

- [x] `npm run check:per-rule-budget` → `9/10 ok; 0 over budget; 1 unmeasured`.
- [x] `npm run check:audit-freshness` → `✅ Per-rule p95 budget: fresh (age 7d, TTL 90d)`.
- [x] Toolchain string verified field-by-field against `benchmarks/results/ilb-flagship/2026-09-07.json`.
- [x] Committed through lefthook; no `--no-verify`.

## Summary by CodeRabbit

* **Documentation**
  * Updated the per-rule budget report with results from the 2026-09-14 run.
  * Expanded measured latency coverage to 8 of 10 rules; 1 rule remains unmeasured.
  * Updated budget status summaries, badges, metadata, and source validation details.
  * Recorded the updated Node.js toolchain and dated benchmark results file.


ofri-peretz:
The Gate failure here is not this PR's content — it is a real bug in the build lane's affected computation, and #1038 fixes it.

`package.json` declares four workspace globs (`apps/*`, `packages/*`, `tools/*`, and the literal `benchmarks`); `ci-build.mts` hard-coded three. `decideAffected` resolves `benchmarks/budgets/per-rule-p95.json` to the touched dir `benchmarks`, nothing in the build universe could claim it, and the guard correctly refused rather than reporting a vacuous pass. **Any** benchmarks-only PR hits this.

`ci-shard-affected.mts` already documents the same hole being fixed for the *test* lane; the build lane never got the same treatment. Rebasing this once #1038 lands.

ofri-peretz:
#1038 landed, so the Gate bug that was blocking this is gone and the branch is updated.

One note on ordering: #1036 adds `benchmarks/results/ilb-flagship/2026-09-14.json`. Its pruning keeps the three most recent runs (08-31, 09-07, 09-14), so the `2026-09-07.json` this PR cites **survives** — the citation and `lastValidated` stay true either way. If #1036 lands first I'll revalidate against 09-14 so the budget points at the newest run rather than one behind.

PR #1037, opened 2026-09-14
fix(rules): FP/FN sweep 2026-09-14 — 6 findings

ofri-peretz:
Six confirmed rule defects found by running every Interlace plugin over the
`burgee` corpus, each one minimized, reproduced mechanically, and then argued
against by a second, independent agent that was given only the snippet and the
rule's docs. Only findings that agent independently agreed with are here.

Sweep: 1,879 findings across 143 files, 0 parse errors. Re-baselined against the
current corpus after the fixes: **1,878 findings, no rule newly firing.**

---

## 1. `node-security/no-shell-injection` — FN (CWE-78, CVSS 9.8)

**Surfaced by:** docs-grounded. Burgee's only two `shell: true` sites are
vendored tests, which `skipTestFiles` correctly excludes.

```js
import { spawn } from 'node:child_process';
spawn(`tar -xzf ${archivePath}`, { shell: true });
```

**Contract violated** — `docs/rules/no-shell-injection.md:43-45`, under
*Examples of incorrect code*:

> ```js
> // A shell is a shell even when you asked for spawn.
> spawn(`tar -xzf ${archivePath}`, { shell: true });
> ```

and `:76`, in *The fix is always the same shape*:

> `spawn(\`cmd ${arg}\`, { shell: true })    -> spawn('cmd', [arg], { shell: false })`

**Cause:** `SHELL_EXEC_FUNCTIONS` was `{exec, execSync}`, so the spawn family
never reached the shape check. Verified against **all 43 rules in the plugin** —
only an unrelated lock-file diagnostic fired. Nothing caught it.

**Fix:** admit `spawn`/`spawnSync`/`execFile`/`execFileSync` **only** when an
options object carries a statically truthy `shell`. The key is read through
`objectKeyName` (so `{'shell': true}` counts) and every argument is scanned
(`execFile`'s options index varies). Existing module-evidence, string-shape and
constant-folding gates are untouched — that is what keeps the safe argv form
silent. With no `shell` the rule stays silent: an interpolated *program name* is
CWE-114, owned by `detect-child-process`, and reporting it here would recreate a
false-positive class this rule already paid for.

Fixtures failed before, pass after.

## 2. `secure-coding/no-improper-type-validation` — FP

**Surfaced by:** `packages/burgee/src/commander/command.ts:538`

```ts
if ((payload as object) !== null && typeof payload === 'object') { go(); }
```

**Contract violated** — `docs/rules/no-improper-type-validation.md:88-91` prints
that exact condition, minus the cast, under `### ✅ Correct`.

**Cause:** `sameExpression` compared operands by node type, so a `TSAsExpression`
never equalled the bare `Identifier` and the null guard vanished. Seven shapes
reproduce — `as`, `satisfies`, `<T>x`, `x!`, chained casts, casts inside a member
chain, and casts on either side of the pair.

**Fix:** unwrap type-only wrappers recursively before comparing. These nodes emit
nothing, so this cannot lose a true positive. The file already argues the
principle: a pinned test one block above records that *"a verdict that turns on
the spelling of a no-op clause is not evidence."*

**Corpus effect:** 29 → 28 findings; no other rule moved.

## 3 & 4. `reliability/no-await-in-loop` — FP + duplicate diagnostics

**Surfaced by:** `scripts/rank-dependents.ts:424` (reported, and reported twice)

```ts
for (const p of await Promise.all(names.map(probe))) record(p);
```

**Contract violated** — the rule's harm is per-iteration serialization
(`docs/rules/no-await-in-loop.md:72-73`, *"N*latency instead of 1x"*). A for-of's
iterable is evaluated **once**, before iteration. Worse, the reported expression
is the `Promise.all` remedy the docs prescribe at `:185`, and the rule offers no
fix or suggestion — the only escape was a disable comment.

**Fix:** exempt only the genuinely once-evaluated slots — `for-of`/`for-in`
`right`, `for` `init` — and attribute each await to its innermost enclosing loop.

⚠️ The obvious fix here is wrong, and the verification gate caught it. Scanning
`node.body` only would have created **four false negatives**: a `for` test and
update, and `while`/`do-while` tests, all re-run every iteration —
`while (await hasMore())` is canonical sequential pagination. Each is now pinned
as an invalid fixture, as is a `for-of` binding default.

The two fixes also interact: naively combined they cancel into a false negative,
because a once-evaluated head inside an outer loop still costs Nx. It is
attributed to the **enclosing** loop, and that case is pinned too.

## 5. `react-features/require-render-return` — FN

**Surfaced by:** docs-grounded — `apps/docs` has no class components.

```tsx
class AnotherComponent extends React.Component {
  render() {
    if (this.props.show) { return <div>Content</div>; }
    // Missing return for else case
  }
}
```

**Contract violated** — `docs/rules/require-render-return.md:53-60` prints this
verbatim under `### ❌ Incorrect`, paired at `:72-79` with a `### ✅ Correct`
block of the *same class name* differing by one line annotated
`return null;  // Explicit return`.

**Cause:** `checkStatement` asked whether *some* path returns, not *every* path:
`hasReturnStatement(consequent) || (alternate && ...)`.

**Fix:** an `if` counts as returning only with an `else` where both branches
return. Trailing returns and `if`/`else`-both-return stay valid.

**One pre-existing test moved from `valid` to `invalid`:**
`coverage-gaps-3.test.ts:738-741` asserted `if (x) return null;` with no else to
be silent. That pinned the defect rather than an intent — the file is
coverage-driven characterization whose cases are named after the implementation
lines they exercise (`(L61 false, L71)`), and neither `KNOWN-LIMITATIONS.md` nor
the rule doc's own *Known False Negatives* section mentions the shape.

## 6. `conventions/no-console-spaces` — fixer emits unparseable code

**Surfaced by:** `scripts/run-evals.ts:265`, `:285`, `:293`,
`scripts/control-bands.ts:574`, `scripts/brand.mts:516` — the five sites this
fixer rewrites. None happens to contain an apostrophe, so burgee survived by luck.

```js
console.log("it's here ");   // --fix -> console.log('it's here');   parse error
console.log("a\nb ");        // --fix -> unterminated string literal
console.log('a\\b ');        // --fix -> value silently becomes `a` + U+0008
```

**Contract violated** — `docs/rules/no-console-spaces.md:23`:
`| **Auto-Fix** | ✅ Yes (removes extra spaces) |`. It removes spaces and also
destroys the file.

**ESLint does not protect the user here.** `verifyAndFix` returns `fixed: true`
with the broken text and the CLI writes it to disk; the multipass loop merely
stops at the next parse. The backslash case emits no diagnostic at all.

**Cause:** the fixer spliced `staticString()`'s **cooked** value — escapes already
resolved — between single quotes with no re-escaping. The author had already met
this hazard from the other side: the `TemplateLiteral` branch is excluded with a
comment saying re-emitting as `'...'` would convert every backtick to a quote.

**Fix:** `toSingleQuoted()` re-escapes backslash, quote and line terminators.
Single quotes are kept because eight existing fixtures pin that style.

---

## Verification

Every finding: minimized to a standalone snippet, reproduced in `RuleTester`,
then given to a **separate** agent with only the snippet and the rule's docs and
asked to argue the report was wrong. Only `AGREE-DEFECT` verdicts shipped. Each
fixture was confirmed **failing before** the fix and passing after, and every
pre-existing test still passes.

That gate did real work beyond confirming: it killed the naive `no-await-in-loop`
fix before it shipped four FNs, and it corrected the `no-improper-type-validation`
report's claim that `null`/arrays "provably cannot reach" an `instanceof` body
(false — `Symbol.hasInstance`, and `[] instanceof Object`), so that finding rests
on the docs' own ✅ example instead.

Gates: `ci:local` green, plus the `quality.yml` battery lefthook does not run —
`check:spellings`, `check:rule-cases`, `check:computed-keys`, `lint:taxonomy`,
`check:key-vocabulary`, `check:name-vocabulary`, `check:case-registry`,
`rule-audit:gate`, `check:format-drift`, `changeset:status`, `lint:publish`,
`check:corpus-coverage`, `check:benchmark-results-size`, `lint:workflows`,
`audit:portability:ci`, `check-links`.

## Screened and rejected

Recorded so future runs do not re-litigate them:

- **`react-features/jsx-no-bind` arrows** — docs promise arrow detection, but
  `KNOWN-LIMITATIONS.md:153-155` already declares it a known FN and
  `performance/react-no-inline-functions` covers it. Internal docs contradiction;
  doc-side fix.
- **`secure-coding/no-redos-vulnerable-regex`** — scslre's `Move` class is
  disabled with no comment and no doc mention; 10 burgee patterns silent,
  `/\s*$/` measured cleanly quadratic (3.5 s at 80k chars). Arguably disclosed by
  the docs' wording → docs gap. Do **not** simply enable `Move`: `/(a|b)+c/` is a
  pinned valid fixture that scslre reports as `Move`.
- **`browser-security/no-sensitive-localstorage`** — the shape is listed both as
  an ❌ example and as a declared known FN. Self-contradicting doc.
- **`reliability/no-unhandled-promise` @ `closeout/registry.ts:120`** — a prior
  deferred finding that no longer reproduces; the site was refactored.
- **`maintainability/nested-complexity-hotspots`, `no-missing-error-context`** —
  both verified real in `RuleTester` (depth walk crosses function boundaries;
  `AggregateError`'s message is argument 2) but **zero** burgee anchors, so not
  eligible.
- **`import-next` true negatives** — `no-duplicates`, `enforce-import-order`,
  `group-exports`, `prefer-node-protocol`, `no-commonjs`, `no-amd` and 8 more
  each grepped against their documented pattern: genuinely conformant.
- **99 rules across node-security / browser-security / modernization /
  operability** screened by linting each rule's own ❌ examples: 82 of 102 blocks
  reported, and the 20 silent ones were nearly all documented partitions to a
  sibling rule that *does* report.

## Confirmed but over the 6-finding cap — queued for the next run

Highest-value first; all recorded in `seen.json`:

1. **`import-next/consistent-type-specifier-style`** — under
   `verbatimModuleSyntax` the `prefer-inline` fixer turns an erased type import
   into a runtime module load. Proven end-to-end with `tsc` + `node`: the fixed
   output throws `ReferenceError: Cannot access 'builtins' before initialization`.
   The `typescript` preset enables this fixer *and* `no-cycle` together, so it
   manufactures the very cycle `no-cycle` promises to catch.
2. **`import-next/no-cycle`** — inert on burgee (142 files, 3 real cycles, 0
   reports): the resolver cannot map NodeNext `.js` specifiers to `.ts` sources.
3. **`react-features/display-name`** — anonymous HOC-returned components never
   visited. Note for the fix: a blanket `ReturnStatement` visitor over-reports on
   render-prop factories, which never mount.
4. **`conventions/consistent-existence-index-check`** — the fix is withheld by
   syntactic position, a boundary absent from the source's own "THREE boundaries"
   list; 23 burgee sites → 9 fixed / 14 unfixable, leaving 3 files carrying both
   spellings after `--fix`.
5. **`reliability/require-network-timeout`** — hardcodes `arguments[1]` for every
   axios call; verified against axios 1.18.1 types, config is index 2 for
   `post`/`put`/`patch` and 0 for `create`/`request`.
6. **`maintainability/no-silent-errors`** — `allowWithComment` only accepts
   comments *above* the catch, never inside it, which is where all four burgee
   sites put them.

## Environment note for the maintainer

The local checkout's `node_modules` was stale against the lockfile and two gates
were failing for that reason alone, not because of any code here:

- `recheck` (a declared devDependency of `eslint-plugin-secure-coding`) was not
  installed, failing 6 `redos-oracle` tests and holding the package below its
  100% coverage threshold.
- `@typescript-eslint/parser` was **8.54.0** installed vs **8.68.0** pinned, which
  makes `utm-taxonomy.ts:156`'s `cooked ?? raw` fallback unreachable and held
  `eslint-plugin-conventions` at 99.85% branches.

`npm ci` in the worktree resolved both (secure-coding → 100%/100%, conventions →
694/694 branches). Worth an `npm ci` in the main checkout.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved `no-await-in-loop` to avoid false positives for awaits evaluated only once and prevent duplicate reports.
  - Fixed `no-console-spaces` autofixes so apostrophes, newlines, Unicode separators, and backslashes remain valid and preserve string values.
  - Improved `no-improper-type-validation` handling of TypeScript type-only casts.
  - Strengthened `require-render-return` to detect fall-through paths.
  - Expanded `no-shell-injection` detection for shell-enabled process execution and dynamic arguments.
- **Documentation**
  - Updated rule release notes and tracking data for the fixes.
- **Tests**
  - Added regression coverage for the corrected rule behavior.


ofri-peretz:
Both `no-shell-injection` findings were real. Fixed in 5beee7ec2, each reproduced with RuleTester **before** the change.

**Backtick shell path** — `hasTruthyShellOption` accepted only `Literal`, so `{ shell: \`/bin/bash\` }` returned false and suppressed the report entirely:

```
execFile(\`ls \${userInput}\`, { shell: \`/bin/bash\` })
→ before: Should have 1 error but had 0
```

**Args array under a shell** — node joins argv into the single string it hands the shell, unescaped, and the rule read only `arguments[0]`:

```
spawn('git', [\`clone \${userRepo}\`], { shell: true })
→ before: Should have 1 error but had 0
```

The inverse is the half that mattered more: with **no** shell, node execs the binary directly and the element is one opaque argument. Inspecting args unconditionally would have turned the *safe* spelling into an FP — worse than the FN. So the check is gated on a truthy shell, and `spawn('git', [\`clone \${userRepo}\`])` is locked as `valid`. Same for an empty and a dynamic backtick shell.

87 → 89 tests; package suite 2991 passing with its 100% branch gate green. The emptiness test reads `raw` rather than `cooked` because `cooked` is nullable only for tagged templates — a branch that can never take its other side, which that gate rejects while tsc still demands it be handled.

Remaining three threads not yet addressed: the `require-render-return` switch analysis (flagged Heavy lift), the `AST_NODE_TYPES` constants in `no-improper-type-validation`, and the `\\r`/U+2028/U+2029 coverage gap in `no-console-spaces`.

ofri-peretz:
Both remaining small findings are in as c0da57532.

**`no-console-spaces` escape paths** — the finding came with its own falsification, so I ran it rather than taking it on trust. Deleting `.replace(/\\r/g, …)` and the two line-separator replacements leaves **all 56 existing tests green** while the fixer emits an unterminated string literal for each value. Both U+2028 and U+2029 are ECMAScript line terminators, so the fixed line stops parsing.

Three fixtures added, then the same mutation re-run against them:

```
× a carriage return is re-escaped so the fixed line still parses
× U+2028 is re-escaped — it is a line terminator, not a space
× U+2029 is re-escaped — it is a line terminator, not a space
Tests  3 failed | 56 passed (59)
```

Exactly the three new cases die and nothing else moves — the mutant is killed, not merely covered.

**`no-improper-type-validation` node types** — correct, and it is that package's own `AGENTS.md:27`. All four names exist in the devkit's curated re-export (the one that deliberately avoids pulling a 24 MB compiler). 135 tests pass.

Both edited rules paid the description ratchet, which is what it is for. 14790 → 14789.

That leaves **one thread open**: the `require-render-return` path-sensitive switch analysis. It is flagged Heavy lift and it is a genuine semantics decision about how far the rule should model control flow — I have deliberately not guessed at it.

ofri-peretz:
Done in baf93aa76 — the finding was right, and I confirmed the FN with RuleTester before touching the rule (`Should have 1 error but had 0` for both shapes below).

```js
switch (k) { case 1: return <A/>; }              // unmatched k escapes
switch (k) { case 1: break; default: return X; } // case 1 breaks out
```

A switch now counts as returning only when a `default` exists and every clause ends in a return, walked from the last clause upward. An empty consequent is deliberate fallthrough and inherits the clause below it; one ending in `break` exits without returning. Clause bodies go through `checkStatement`, so an `if`/`else` returning on both arms counts — **no clause has to carry a bare `return` of its own**, which is the part you asked to preserve.

Six named cases added. Against the previous implementation exactly the three invalid ones fail and all three valid ones pass, so this adds no false positive on code the rule already accepted. 23 tests, up from 17.

One thing worth flagging: an existing case in `coverage-gaps-3.test.ts` had to move from `valid` to `invalid` —

```js
render() { switch (x) { case 1: log(); break; default: return null; } }
```

`x === 1` leaves render() returning undefined. It was written to reach a coverage line in the old any-clause-returns check, and that check is the only reason it passed. The `if` case directly below it was moved out of `valid` by an earlier sweep for exactly the same reason, with the same note — so this is the second time a coverage-shaped fixture had pinned this rule's false negative.

